PR #1, opened 2020-10-05
Add event hooks to client

nmuldavin:
* Add init hook, exposed through postmate model. Must be called by parent after handshake with app context data before all other operations may proceed.
* Add exec hook, exposed through postmate model, for the parent to send events to the child.
* Add `on` event handler subscription method, with subscribe and unsubscribe logic
* Add checks in `exec` handler verifying that the user installation includes the capabilities necessary to recieve an event of the given type. If not, a debug error is logged and the subscribed handler is not executed.


enbashi (approved):
looks great! The sandboxing is super helpful.

PR #6, opened 2020-11-04
Add event triggering from child side

enbashi:
- Add support for triggering events from the child side
- Introduce new capabilities and events
- Update `.gitignore` to ignore `dist`
- Breakdown events into two groups based on direction
- Update sandbox tests
- Bump `framepost` to v 0.1.2


sandbox/child/index.html
```diff
@@ -15,12 +16,15 @@ <h4>Datadog UI Apps SDK Sandbox: Child</h4>
             client.on('dashboard_cog_menu_context', context => {
                 console.log('Received dashboard cog menu context', context);
 
-                const a = document.createElement('a');
-                a.href = context.shareToken;
-                a.target = '_blank';
-                a.innerHTML = 'Link';
```
enbashi, line 21: this no longer works with `sandbox` attr.

package.json
```diff
@@ -17,7 +17,7 @@
         "format": "prettier --write .",
         "prepare": "yarn build",
         "sandbox": "node scripts/sandbox.js",
-        "start": "yarn watch & webpack-dev-server --mode development --progress --color --hot",
+        "start": "yarn watch & webpack serve --mode development --progress --color --hot",
```
enbashi, line 20: https://github.com/webpack/webpack-dev-server/issues/2029

sandbox/child/index.html
nmuldavin: See my comment below

src/capabilites/index.ts
```diff
@@ -1,16 +1,52 @@
...
     }
 }
 
-export const capabilityManagers = [DashboardCogMenuManager];
+class AppRoutingManager extends CapabilityManager {
+    type = UiAppCapabilityType.APP_ROUTING;
+    eventsToSubscribe = [];
+    eventsToTrigger = [
+        UiAppEventToTriggerType.RELOAD_FRAME,
+        UiAppEventToTriggerType.OPEN_URL
```
nmuldavin, line 26: I'm thinking maybe we just add an `allow-popups` sandbox attribute to the iframes so that they can contain `target="_blank"` links? That seems simpler than requiring that they use an event. It could potentially mean some apps could get annoying with many popups, but the user can always uninstall. I don't think there's a security risk. I definitely see a need for a different `navigate` event, in which the iframe requests that the entire app navigate to a new page (not in a new tab). The usecase would be for programatically navigating somewhere else within the datadog app, like to a certain infrastructure view or to a custom page when we add that capability

nmuldavin (commented):
Code looks great, the split of `events` to `eventsToSubscribe` and `eventsToTrigger` makes a lot of sense. I left a comment for you re: the `open_url` event, let me know what you think

src/capabilites/index.ts
enbashi: that was my first approach actually, but it turns out even with `allow-popups`, the new page will have limitation, for example you can't submit forms. This approach brings the control back to datadog which allows us to expand on it in the future to allow white/black listing of urls and can evolve into navigating routes within the same app or different apps. This made more sense to me.

nmuldavin: We should definitely enable `allow-forms`, that should handle the case you're mentioning. Anyways, thanks for explaining. We'll see how it goes,, if it turns out to be a huge hassle for developers we can turn it on later

nmuldavin (approved):

PR #11, opened 2020-12-03
Named api methods

nmuldavin:
In addition to the existing generic api methods added in #9, this creates named methods for each of the GET metrics endpoints:

```
client.api.v1.metrics.getMetadata(metricId);
client.api.v1.metrics.listActiveMetrics({ from, host });
client.api.v1.metrics.search({ q: '...' });
client.api.v1.metrics.query({ from, to, query });
```

I also added appropriate unit tests since I neglected those in the last PR

### DIFF
--- a/src/api/api.test.ts
+++ b/src/api/api.test.ts
@@ -1,0 +1,149 @@
+import {
+    IFrameApiRequestMethod,
+    UiAppRequestType,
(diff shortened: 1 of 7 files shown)


enbashi (approved):
Looks great!

PR #13, opened 2020-12-07
[app platform] Add dashboard context menu feature

DrkSephy:
This pull request adds a new feature to the SDK, `dashboard_context_menu`.

### DIFF
--- a/src/features/index.ts
+++ b/src/features/index.ts
@@ -1,10 +1,12 @@
 import { dashboardCogMenu } from './dashboard-cog-menu';
+import { dashboardContextMenu } from './dashboard-context-menu';
 import { dashboardCustomWidget } from './dashboard-custom-widget';
 import { modals } from './modals';
 import { sidePanels } from './side-panels';
 
 export const features = [
     dashboardCogMenu,
+    dashboardContextMenu,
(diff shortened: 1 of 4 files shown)


src/features/index.ts
```diff
@@ -8,6 +8,16 @@ import { FeatureManager } from './featureManager';
 
 export type { FeatureManager } from './featureManager';
 
+class DashboardContextMenuManager extends FeatureManager {
+    type = UiAppFeatureType.DASHBOARD_CONTEXT_MENU;
+    eventsToSubscribe = [UiAppEventToSubscribeType.DASHBOARD_CONTEXT_MENU];
```
nmuldavin, line 13: What does this event do?

DrkSephy: Before the recent changes, it would inform the 3p app when the context menu is triggered -- this may be outdated now.

DrkSephy: Another note: this event doesn't send any new information to the 3p app as it is still TBD what type of context would be useful.

nmuldavin: thanks for the context!

enbashi (approved):

PR #15, opened 2020-12-08
Custom Events

nmuldavin:
At a top level this PR introduces custom event broadcasting / subscription with `client.events.broadcast()` and `client.events.onCustom()`

It also introduces some major changes / refactors, in general to move us towards a more modular architecture in this sdk where there will be many namespaces on the main sdk client handling different sorts of things:

* `client.api` -> datadog API stuff
* `client.events` -> events stuff
* There will be more (e.g. client.modal, client.navigation)

Crucially, these 'modules' are not necessarily related 1-1 to features. the api module, for example, has nothing to do with features at all. To implement this PR:
* Move event handling logic to a `DDEventsClient` instance namespaced at `client.events`:
  * `client.on()` -> `client.events.on()`
  * `client.onCustomEvent()` -> `client.events.onCustom()`
* Remove class-based `FeatureManager`s in favor of static feature config
* Unify enums `UiAppEventToSubscribeType` and `UiAppEventToTriggerType` in to single `UiAppEventType`. This is for two reasons:
  1. To maintain consistency with web-ui typing
  1. We will favor use of `framepostClient.request()` over `framepostClient.send()` for child -> parent communication, so that the parent can notify the child of the success / failure of the requested operation. Events that we would have put in `UiAppEventsToTriggerType` we will move to `UiAppRequestTypes`
* remove `client.triggerEvent()` method, in favor of `client.broadcast`
* temporarily remove app routing feature and related events. this will be added back in a subsequent PR


enbashi (approved):
🔥

PR #16, opened 2020-12-17
Type Bugfixes

nmuldavin:
* Align AppContext type with web-ui update
* Point package.json types do `dist/index.d.ts`

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -18,13 +18,18 @@ export interface HandleEventParams<T = any> {
 }
 
 export interface AppContext {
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -18,13 +18,18 @@ export interface HandleEventParams<T = any> {
 }
 
 export interface AppContext {
-    // Name of current user
```
iddl, line 21: you also need to add `frameContext.options` as a generic object I guess

iddl (approved):

PR #22, opened 2021-01-26
Add optional context for side panel

enbashi:
Support an optional second argument for `client.sidePanel.open()` which is a custom context than can be passed from the sdk.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -75,17 +75,14 @@ export interface DashboardWidgetContext {
 export interface MenuItemContext {
     key: string;
 }
-// Generic type of optional arguements passed to different feature components like modal, sidepanel, etc
-export interface CustomFeatureContext {
-    [key: string]: any;
-}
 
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -75,12 +75,17 @@ export interface DashboardWidgetContext {
...
+export interface CustomFeatureContext {
+    [key: string]: any;
+}
 
 // A combined object specifing data about the context of a feature within the app
 export interface FeatureContext {
     dashboard?: DashboardContext;
     widget?: DashboardWidgetContext;
     menuItem?: MenuItemContext;
+    sidePanel?: CustomFeatureContext;
```
nmuldavin, line 88: Do you need to update this to match the change in web-ui?

enbashi: ah I made the change in `/dist` instead :). Thanks for catching this. Fixed.

nmuldavin (approved):
LGTM!

PR #26, opened 2021-02-08
Dynamic menu items

nmuldavin:
* Adds `dashboardCogMenu.onRequestItem()` for supplying items dynamically
* Renames `dashboardCogMenu -> widgetContextMenu()`
* Adds `widgetContextMenu.onRequestItem()` for supplying items dynamically
* Remove support for key-based open-modal and sidepanel calls (breaking change)


src/widget-context-menu/widget-context-menu.ts
```diff
@@ -0,0 +1,75 @@
...
+        );
+
+        // initialize with an empty reponse handler
+        this.onRequestItems(() => emptyConfig);
+    }
+
+    /**
+     * Registers a request handler for providing context menu items dynamically
+     */
+    onRequestItems(
```
enbashi, line 30: while technically `onRequestItems` makes perfect sense to describe what this method does, I think either replacing it or wrapping it with something more generic like `init` make more sense and could be more intuitive to users. This allows us to future proof it so it be reused it if we decide to allow users to define other properties in addition to items like titles for example. We could also use the same name for more features in the future that are not necessarily items. What do you think?

enbashi (approved):
Looks great. I added a couple of suggestions. None is a blocker.

src/types.ts
```diff
@@ -117,6 +121,27 @@ export interface IFrameApiRequest<Q> {
...
+}
+
+export interface EventMenuItem extends MenuItemCommon {
+    type: MenuItemType.EVENT;
+}
+
+export type MenuItem = LinkMenuItem | EventMenuItem;
+
+export interface MenuItemRequestResponse {
+    items: MenuItem[];
```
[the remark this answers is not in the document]
nmuldavin: I don't have a specific usecase in mind, I guess this was motivated by experience with API design (since this is an API of sorts). If your API returns an array and it turns out you need some other data about the array later, you'd have to make a breaking API change. If it's an object, and you need some new piece of data you just add a key

src/widget-context-menu/widget-context-menu.ts
nmuldavin: We have a pattern already of prefixing with `on` when registering event or request handlers, so i think we should use `on`, but I think renaming it to `onRequest` makes sense and is shorter. Does that make sense to you?

enbashi: yeah sure, i think that's a good middle ground :)

PR #28, opened 2021-02-12
Add secrets.remove and support for OAuth flow 

enbashi:
Add the following public methods:
- `client.secrets.requestAuthTokens()`
- `client.resolveAuthTokens()`
- `client.secrets.remove()`


nmuldavin (approved):
Nice work. As we discussed in slack i think it's possible we want to rename these methods depending on where the secrets / auth feature work goes, but lets get this in we can figure it out

PR #29, opened 2021-02-23
Misc. Development experience updates

nmuldavin:
* Supports new context_change event, sent from the parent whenever the frame context changes. The event can be subscribed to
* Uses ^ to update data returned from client.getContext() so that it is always current
* Upgrade framepost to take advantage of error handling updates in https://github.com/DataDog/framepost/pull/6
* Turn on debug mode based on context data sent from parent. This allows debug mode to be turned on at runtime, by the user in web-ui, rather than by the developer at build time
* Turns off debug mode in framepost. This optimizes for 3p developer experience, since they will not likely care about framepost internals. We can add a 'master of the universe' mode for DD ui apps devs later if we need it
* Prefix logs with the current url. This allows 3p devs to tell which iframes the debug messages are coming from

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
         "access": "public"
     },
     "dependencies": {
-        "@datadog/framepost": "https://github.com/DataDog/framepost.git#263c75d619ada57833e0cb6f0d7de386a3ccc136"
+        "@datadog/framepost": "0.2.2"
     },
     "devDependencies": {
         "@types/jest": "^26.0.14",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
         "access": "public"
     },
     "dependencies": {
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
         "access": "public"
     },
     "dependencies": {
-        "@datadog/framepost": "0.2.0"
+        "@datadog/framepost": "https://github.com/DataDog/framepost.git#263c75d619ada57833e0cb6f0d7de386a3ccc136"
```
enbashi, line 31: did you mean to check this in? I approved the framepost change if you want to just bump the version instead.

enbashi (approved):

package.json
nmuldavin: I did temporarily, makes it easier for you to test without needing to link everything. Changing to the real version now

PR #31, opened 2021-03-22
[requests] Add DashboardClient for sending events from SDK to parent dashboard

hydrosquall:
## Motivation

- Enable UI App developers to trigger state changes in the parent Dashboard (Cursor and Timeframe)

## Changes

- Modeled after the Secrets / Location clients
- Didn't name this the `requests` client  since it's not the only client making requests, but open to renaming. To me, an `action` is something the child asks a parent dashboard to do. (This could also be a `DashboardActions` client depending on whether widgets may live on pages that aren't dashboards someday and need to operate with the same API, or they'd behave differently).

## Testing

- See https://github.com/DataDog/web-ui/pull/30273 for corresponding PR with listeners

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -10,7 +10,7 @@ export class DDActionsClient {
         this.client = client;
     }
 
-    async setDashboardCursorRequest({ timestamp }: SetDashboardCursorRequest) {
+    async setDashboardCursor({ timestamp }: SetDashboardCursorRequest) {
         return this.client.framePostClient.request<SetDashboardCursorRequest>(
             UiAppRequestType.SET_DASHBOARD_CURSOR,
             {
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -17,7 +17,14 @@
         "esModuleInterop": true,
         "resolveJsonModule": true,
         "sourceMap": true,
-        "skipLibCheck": true
+        "skipLibCheck": true,
```
hydrosquall, line 20: My IDE was complaining about the usage of `Set` when `lib` wasn't included.

src/actions/actions.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    async setDashboardTemplateVars({ vars }: SetDashboardTemplateVarsRequest) {
+        return this.client.framePostClient.request<SetDashboardTemplateVarsRequest>(
+            UiAppRequestType.SET_DASHBOARD_TEMPLATE_VARS,
+            {
+                vars
+            }
+        );
+    }
+}
```
nmuldavin, line 39: These all make sense but i'd say we should call it a `DDDashboardClient` and have the methods be `client.dashboard.setCurosor()`, `client.dashboard.setTimeframe()`, `client.dashboard.setTemplateVars()`

nmuldavin (commented):
This looks good but there are linting errors that didn't get caught because github actions didn't run. Can you resubmit? I added a new trigger type for the ci workflow that i think will work

nmuldavin (approved):
LGTM!

PR #37, opened 2021-04-26
[UI Apps] Conform FramePost client handshake timeout to ParentClient in web-ui

wfa207:
## Motivation

To conform other usages of FramePost, we're increasing the request timeout in the SDK's FramePost client.

## Changes

This change increases the handshake timeout to **10,000ms**. I've also abstracted the FramePost settings to a separate constant.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -49,8 +49,8 @@ export enum IFrameApiRequestMethod {
 export const FramePostClientSettings = Object.freeze({
     // 3p devs most likely dont need to see framepost debug messages
     DEBUG: false,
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -46,6 +46,14 @@ export enum IFrameApiRequestMethod {
     DELETE = 'DELETE'
 }
 
+export const FramePostClientSettings = Object.freeze({
```
nmuldavin, line 49: 🧊

nmuldavin (approved):

PR #45, opened 2021-08-17
Remove identifying info from context

nmuldavin:
remove identifying info from app context

### DIFF
--- a/sandbox/parent/index.html
+++ b/sandbox/parent/index.html
@@ -22,12 +22,6 @@
 
             frame.onload = () => {
                 client.requestChannel(frame, {
-                    name: 'User',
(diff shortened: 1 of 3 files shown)


joneshf-dd (commented):
I noticed there's a reference to user stuff in the parent sandbox: https://github.com/DataDog/ui_apps_sdk/blob/d2f80abdadb7012a81f7fb3f59f0af86200e4d41/sandbox/parent/index.html#L25-L30. Do we need to do something with that as well?

nmuldavin:
Also a good catch. Updated!

joneshf-dd (approved):
Looks good!

PR #47, opened 2021-08-27
Add optional arguments when opening modal

joneshf-dd:
Per request, we want modals to work more like side panels where we allow arbitrary data to be passed in when opening. This is similar to #22.

Docs PR: https://github.com/DataDog/apps/pull/13


enbashi (approved):

enbashi (approved):

joneshf-dd:
Thanks for the reviews!

PR #51, opened 2021-09-07
Make `AuthState` generic

joneshf-dd:
We make it so the `args` of the `AuthState` is now generic and can be decided when initializing the client. This should allow for greater type safety when working with the SDK client and using auth in particular.

Probably the thing to check is that we're okay with the tests added in `index.test.ts`. The reason behind them is explained a bit in the commit, but for posterity, it took a non-trivial amount of time to figure out why things weren't inferring properly, and it was because one of the `DDClient`'s didn't have its type variable properly set (it was defaulting to `DDClient<unknown>`). If we'd prefer a different way, more than happy to switch to that. But this is what I came up with.

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -2,18 +2,21 @@ import type { DDClient } from '../client/client';
 import { RequestType } from '../constants';
 import { AuthState, AuthStateOptions } from '../types';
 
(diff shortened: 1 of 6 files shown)


nmuldavin (approved):

joneshf-dd:
Thanks for the review!

PR #52, opened 2021-09-08
Remove some `any`s from custom widgets

joneshf-dd:
We change how we define custom widget options so they can be typed a bit stricter and remove some `any`s in the process.

Docs PR: https://github.com/DataDog/apps/pull/16


enbashi (approved):

joneshf-dd:
Thanks for the review!

PR #53, opened 2021-09-08
Remove more `any`s

joneshf-dd:
This removes some more `any`s from the SDK. See the commit messages for explanations on why we're using `unknown` instead of generics.


enbashi (approved):

joneshf-dd:
Thanks for the review!

PR #55, opened 2021-09-17
Support auto sizing modal height

joneshf-dd:
We make a new event for resizing the iframe and send it over the wire whenever the iframe is resized. This allows us to set the modal height to whatever the height of the actual content is.

Docs PR: https://github.com/DataDog/apps/pull/24.

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -105,10 +105,12 @@ export class DDClient<AuthStateArgs = unknown> {
     /**
      * Notify the parent that the iframe should be resized.
      *
+     * Will default any dimensions not given to the iframe's actual dimensions.
+     *
      * There's no guarantee that the parent will adjust the iframe's dimensions.
      * The parent will also sanitize the dimensions attempting to keep the iframe within the viewport.
      */
-    resize() {
+    resize(dimensions?: Partial<IFrameDimensions>) {
         const style = window.getComputedStyle(document.documentElement);
         const parsedHeight = parseFloat(style.getPropertyValue('height'));
         const parsedWidth = parseFloat(style.getPropertyValue('width'));
@@ -118,12 +120,12 @@ export class DDClient<AuthStateArgs = unknown> {
         const width = Number.isNaN(parsedWidth)
             ? document.documentElement.scrollWidth
             : parsedWidth;
-        const dimensions: IFrameDimensions = {
-            height,
-            width
+        const computedDimensions: IFrameDimensions = {
+            height: dimensions?.height ?? height,
+            width: dimensions?.width ?? width
         };
 
-        this.framePostClient.send(EventType.RESIZE_IFRAME, dimensions);
+        this.framePostClient.send(EventType.RESIZE_IFRAME, computedDimensions);
     }
 
     /**
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -98,10 +98,20 @@ export class DDClient<AuthStateArgs = unknown> {
         // Since computing the size of an iframe is hard to do correctly,
         // we listen to `'resize'` events so we can send the actual values over.
         window.addEventListener('resize', () => {
+            const style = window.getComputedStyle(document.documentElement);
+            const parsedHeight = parseFloat(style.getPropertyValue('height'));
+            const parsedWidth = parseFloat(style.getPropertyValue('width'));
+            const height = Number.isNaN(parsedHeight)
+                ? document.documentElement.scrollHeight
+                : parsedHeight;
+            const width = Number.isNaN(parsedWidth)
+                ? document.documentElement.scrollWidth
+                : parsedWidth;
             const dimensions: IFrameDimensions = {
-                height: document.documentElement.scrollHeight,
-                width: document.documentElement.scrollWidth
+                height,
+                width
             };
+
             this.framePostClient.send(EventType.RESIZE_IFRAME, dimensions);
         });
     }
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -269,3 +269,8 @@ export interface APIAccessChangeEvent {
 export interface OrderedItem {
     order?: number;
 }
+
+export interface IFrameDimensions {
+    height: number;
+    width: number;
```
joneshf-dd, line 275: Based on the discussion we had, I decided to go with both height and width since it seems like we might want to use this event for things other than modals. Like, I believe Rami mentioned something about custom widgets.

It's not a requirement that we keep both of these. We only need height for modals. So, if we'd prefer to not expand the API too much, we can drop the width.

src/client/client.ts
```diff
@@ -89,6 +90,20 @@ export class DDClient<AuthStateArgs = unknown> {
...
+    private registerEventListeners() {
+        // Since computing the size of an iframe is hard to do correctly,
+        // we listen to `'resize'` events so we can send the actual values over.
+        window.addEventListener('resize', () => {
+            const dimensions: IFrameDimensions = {
+                height: document.documentElement.scrollHeight,
+                width: document.documentElement.scrollWidth
+            };
+            this.framePostClient.send(EventType.RESIZE_IFRAME, dimensions);
+        });
```
joneshf-dd, line 106: I wasn't entirely sure where to put this, but decided the `DDClient` would be a good start. If there's a better place for this (or a different way to do this) more than happy to change it.

joneshf-dd (commented):
Made a contextual review with comments/questions I had.

enbashi (approved):

joneshf-dd:
Thanks for the review!

nmuldavin:
Ok, this can serve as a review for all three PRs so that it's centralized.

I tested this very thoroughly. Two issues, one probably resolvable, and a question:

1. Resolvable issue: At one point I got myself in an infinite loop where the window.resize event in the iframe triggered sending the RESIZE_IFRAME message to web-ui, which in turn set a new height in the iframe, which triggered the resize event in the iframe. That's not good. It also seems to not happen all the time. It might come down to something as silly as a rounding error in the browser? Anyways, It made me question the window.resize listener. Maybe we just manually send the first resize event? Or, we send it with the window.resize() event listener the first time, but in the registered event handler we unsubscribe the window event such that we guarantee that the loop doesn't recur. 
2. Probably unresolvable issue: I spent a long time debugging why i was still seeing vertical scrollbars despite this seeming to work right mostly:
Screen Shot 2021-09-24 at 2 32 40 PM

It seems to come down to incorrect browser measurements when either a) there is text-wrapping (as seen above) or b) when there is a horizontal scrollbar at the bottom, causing the content to shift upwards by the height of the scrollbar. 

I did some digging, this seems to be a tough problem. We might consider posing this question to someone in dashboards or dataviz who has a lot of experience 'measuring' things in the browser, but we should not block on it for now.

3. Question: I like the model where you can provide both `width` and `height` as overrides, but it makes the existing `size` property on modals kindof unnecessary and confusing. A developer might wonder: what's the difference between the size passed there and the SDK-specified size? Should we remove the original `size` parameter in favor of only using the SDK specified size?

joneshf-dd:
Thanks for really hammering on this stuff! Sorry there are still some issues with it.

> 1. Anyways, It made me question the window.resize listener. Maybe we just manually send the first resize event? Or, we send it with the window.resize() event listener the first time, but in the registered event handler we unsubscribe the window event such that we guarantee that the loop doesn't recur.

That makes sense to me. I think it also makes sense now in the world where the App developer can explicitly request a resize (in case that first automatic resize doesn't work out quite right). This is probably the reason most other App platforms don't automatically resize iframes 😅.

> 2. I did some digging, this seems to be a tough problem. We might consider posing this question to someone in dashboards or dataviz who has a lot of experience 'measuring' things in the browser, but we should not block on it for now.

Sounds good. I'll investigate more on Monday.

> 3. Question: I like the model where you can provide both `width` and `height` as overrides, but it makes the existing `size` property on modals kindof unnecessary and confusing. A developer might wonder: what's the difference between the size passed there and the SDK-specified size? Should we remove the original `size` parameter in favor of only using the SDK specified size?

That's fair. I was thinking of `size` as only the initial size and the `DDClient.resize()`. I think most other App platforms support that kind of thing. I don't have a strong feeling one way or the other, so I'm down to remove it to minimize confusion. Just want note (in case it's forgotten when we implement): we currently ignore the width on the other end, and we should start using it if we remove the `size` property.

On that note, could we track usage data like this? I feel like if we saw that nobody (or a very small percentage) used the `size` property, it'd be a lot easier to decide whether to remove it. Alternatively, if we saw the vast majority using it, we might consider keeping it so as to not break a bunch of App developers.

joneshf-dd:
For posterity, we talked offline about this question a bit more:

> 3. Question: I like the model where you can provide both `width` and `height` as overrides, but it makes the existing `size` property on modals kindof unnecessary and confusing. A developer might wonder: what's the difference between the size passed there and the SDK-specified size? Should we remove the original `size` parameter in favor of only using the SDK specified size?

We're going to stick to both dimensions and also keep the `size` in the `ModalDefinition` since knowing the width tends to help the height get computed properly.

enbashi (approved):

joneshf-dd:
Thanks again for the review!

PR #66, opened 2021-12-13
Explain the Node.js-specific documentation

joneshf-dd:
The Datadog API Reference doesn't work for Apps out of the box.

We note down that it's not a straight-shot and will take a bit of work.
Hopefully, this will be addressed by the API client becoming usable in the browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@ Datadog Apps allow third party developers to extend the native functionality off
 * [Frequently Asked Questions](docs/en/faq.md)
 * [UI-Extension Design Guidelines](docs/en/ui-extensions-design-guidelines.md)
 * [Datadog API Reference](https://docs.datadoghq.com/api/latest/)
-    * Note: the TypeScript examples in the Datadog API Reference and the [`@datadog/datadog-api-client`](https://www.npmjs.com/package/@datadog/datadog-api-client) are currently Node.js-only, while Datadog Apps are browser-only. The API Reference is useful to understand the Datadog API, but the examples will need some conversion for the browser.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,7 @@ Datadog Apps allow third party developers to extend the native functionality off
 * [Frequently Asked Questions](docs/en/faq.md)
 * [UI-Extension Design Guidelines](docs/en/ui-extensions-design-guidelines.md)
 * [Datadog API Reference](https://docs.datadoghq.com/api/latest/)
+    * Note: the TypeScript examples here and the [`@datadog/datadog-api-client`](https://www.npmjs.com/package/@datadog/datadog-api-client) are currently Node.js-only, while Datadog Apps are browser-only. The API Reference is useful to understand the Datadog API, but the examples will need some conversion for the browser.
```
nmuldavin: Maybe also something letting devs know we're working on it, like "We are working on updating the api client package to work seamlessly with the iframe sdk. In the mean-time we recommend accessing the api with the ui extensions sdk directly".Maybe link to an example api call too

joneshf-dd: Down to add the note that we're working on it!

> Maybe link to an example api call too

Is it cool if we do this as a follow-up PR in the form of adding a small example in the `examples` directory? Maybe just something that returns the [list of active metrics](https://docs.datadoghq.com/api/latest/metrics/#get-active-metrics-list)? If it's in the `examples` directory, then we can also at least make sure it stays up to date in CI.

nmuldavin: Sure. I think an api access example would be great. I had a link to [here](https://github.com/DataDog/apps/blob/master/docs/en/programming-model.md#accessing-the-api-through-the-sdk) in my head as a smaller thing

joneshf-dd: Oh sweet! I didn't know that existed. I'll add that.

nmuldavin (approved):
Looks great!

joneshf-dd:
Thanks for the review!

PR #67, opened 2021-12-14
UIAPPS-72 Setup release process with `auto`

joneshf-dd:
## Motivation

- We want to setup a release process similar to what the Dataviz SDK uses.
- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-72

## Changes

This is almost all process and infrastructure changes.
- Updated the node version to 16.3.0 for `lerna`.
- Moved over to using `auto` to handle releases.
- Added a document explaining how the releases work now.
- Added a PR Template to make `auto` a bit easier.
- Updated the GitHub actions to run `auto` instead of the previous workflow.

## Testing

Aside from dealing with CI changes, likely the only thing we can do is merge this PR and deal with whatever else we need to follow-up on.

## Change Type

- [x] `patch`
- [ ] `minor`
- [ ] `major`

<details>
  <br />

  ```bash
  npm install @datadog/ui-extensions-sdk@0.24.1-canary.67.3d2324f.0
  # or 
  yarn add @datadog/ui-extensions-sdk@0.24.1-canary.67.3d2324f.0
  ```
</details>


.github/workflows/publish.yml
```diff
@@ -1,18 +1,23 @@
 name: publish
 on:
-    release:
-        types: [created]
+    - push
```
wfa207, line 3: Is there ever a case where we would want to **manually** trigger publish? If so, is there another action we can use to trigger that? Possibly a tag?

wfa207 (approved):
This LGTM — haven't really worked with `lerna` before to manage nested packages, but seems sensible.

I had one question about whether we want to be able to trigger GH Actions with a tag in addition to merges / pushes, but it's not blocking.

.github/workflows/publish.yml
joneshf-dd: I don't know when we'd want to manually trigger this `publish` workflow, but we can by listening to the [`workflow_dispatch` event](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually). That allows the `publish` workflow to be [run manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow).

Do  you think we should add that now?

wfa207: Eh, if we don't have a reason to yet, then I don't think it's worth spending any time on it. I think knowing that we have a path forward is enough, in case we ever need it.

joneshf-dd: Okay in that case, we can at least document the idea in `RELEASE.md`.

joneshf-dd:
Aight, had to make a few updates to get the canary releases working. Mind giving that another once-over when you get a minute?

RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Canary workflow
+
+When a PR is opened, [`auto`][auto] will create a "canary" release, publish it to `npm`, and report back in the PR how to use the "canary" release.
+The "canary" release can be used to test out some changes without fully merging the PR.
+
+## Manual releases
+
+There is not currently a (documented) manual release process; neither through GitHub Actions nor outside of GitHub Actions.
+If we find that we need/want the ability to manually release through GitHub Actions, we can look into using the [`workflow_dispatch` event][workflow_dispatch] to [manually run a workflow][]
```
wfa207, line 20: Thanks for this!

wfa207 (approved):
Looks good!

joneshf-dd:
Thanks for the reviews!

hydrosquall:
Note from the future: currently lerna is a requirement for using auto with monorepos. I looked briefly into making it work using just Yarn 2 features instead, but didn't stick with it because Lerna worked well enough for our usecase.

https://github.com/intuit/auto/discussions/1861#discussioncomment-1004780

PR #68, opened 2021-12-14
Fix `sentiment` example build

joneshf-dd:
While trying to put together https://github.com/DataDog/apps/pull/67, it looks like the `sentiment` example is breaking due to lint errors. Rather than continue to muddy up that PR, the fix is in this PR instead.

The actual error is in the [first commit's build](https://github.com/DataDog/apps/runs/4514539207?check_suite_focus=true).


nmuldavin (approved):

joneshf-dd:
Thanks for the review!

PR #72, opened 2021-12-17
Add Redis example project

tdimnet:
## Motivation

Ok, it's a huge one!
This is an example project that uses Redis: the idea is to use a Datadog Application to delete Redis keys.

There are some microservices in this project:
- A REST API written in Node that returns products. When the project runs, 10,000 random products are loaded into a MongoDB and a Redis..
- Another REST API written in Python that is used by the Datadog App.
- A Datadog App written in React.
- A bootstrap script that creates a Datadog App and a Dashboard
- A Datadog Agent that enables the Redis integration

I will update the project soon with an example authentication system. I'll see if I create either a new folder with the project or another branch.


joneshf-dd (approved):
Looks good!

PR #73, opened 2021-12-17
UIAPPS-87 Setup `changesets` for package releases

joneshf-dd:
## Motivation

- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-87
- `auto` doesn't really work for us with independent releases. All of the changes end up being patch changes since we couldn't figure out the labels, plus unless we move over to something like conventional commits, every package has to have the same version. That would work in the small, but not if we hope this repo can continue to grow.

## Changes

- We start using [`changesets`](https://github.com/changesets/changesets) for normal package releases. It seems like it'll work better for us: allowing us to version packages independently, yet also allowing some packages to move in [lock step](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#linked-array-of-arrays-of-package-names).

## Releases

Choose one:

- [ ] No release is necessary.
- [x] All packages that need a release have a changeset.

<details>
  <br />

  ```bash
  npm install @datadog/ui-extensions-sdk@0.24.3-canary.73.c88e229.0
  # or 
  yarn add @datadog/ui-extensions-sdk@0.24.3-canary.73.c88e229.0
  ```
</details>


RELEASE.md
```diff
@@ -1,24 +1,21 @@
...
-## Manual releases
+## Package releases
+
+We use [`changesets`][] to help with normal package releases.
 
-There is not currently a (documented) manual release process; neither through GitHub Actions nor outside of GitHub Actions.
-If we find that we need/want the ability to manually release through GitHub Actions, we can look into using the [`workflow_dispatch` event][workflow_dispatch] to [manually run a workflow][]
+[`changesets`][] will manage a long-running PR that batches changes together.
+When a normal PR is merged, [`changsets`][] will keep track of any [changeset][]s in this long-running PR.
+When we're ready to release some number of packages, we can merge this long-running PR and [`changesets`][] will release all packages with a [changeset][].
```
wfa207, line 17: I didn't quite realize that `changesets` manages a long-running PR to control versioning — this being the case, should we change the repository settings to **prevent squash merges**? I imagine that squashing and merging would cause issues in the long-running branch? Maybe I'm misunderstanding

wfa207 (approved):
I had one question about possibly restricting squash merges in the repo based on how I think I understand `changesets` works, but it definitely doesn't block merge.

Thanks for taking charge on this change! This looks promising.

RELEASE.md
joneshf-dd: Far as I understand it, it'll keep the long-running branch on top of the base branch whenever it changes, so it shouldn't matter how we merge into the repo. For example, the `changeset-release/main` branch in https://github.com/changesets/changesets/pull/709 was "rebased<sup>[1]</sup>" at least seven times while more changes were merged into their `main` branch. I think `changesets` uses squash merges (at least that's what it looks like) and it doesn't seem to cause any issues.

Long story short, I don't think it matters. But we can change it if we'd prefer it one way or the other.

<sup>[1]</sup>: I don't actually know if it's rebasing or making a whole new branch locally and force pushing it. It really doesn't matter, the outcome is the same.

joneshf-dd:
Thanks for the review!

RELEASE.md
wfa207: Sounds good! You know I like to Squash & Merge 😂 😭 . If that's the case, we can just leave the settings as-is. If it becomes a problem, it's easy to adjust.

PR #77, opened 2021-12-21
Remove `pull_request_target` from `ci`

joneshf-dd:
## Motivation

- It seems like using the `pull_request_target` in the "Continuous Integration" workflow isn't what we want. For more motivation, see the first commit.

## Changes

- Removes the `pull_request_target` event.

## Releases

Choose one:


nmuldavin (approved):

joneshf-dd:
Thanks for the review!

PR #82, opened 2021-12-21
Explain a more about when to create a changeset

joneshf-dd:
## Motivation

- It's not really clear when to create a `changset`.

## Changes

- We updated the documentation to explain when we likely want a changeset.

## Releases

Choose one:


nmuldavin (approved):

joneshf-dd:
Thanks for the review!

PR #86, opened 2021-12-29
Simplification of stream boostrap:

tdimnet:
## Motivation

Allow the user to launch the Stream example project more easily:
- Remove the dependency installation step.
- Add the `yarn` command in all Dockerfile.


joneshf-dd (approved):
Looks good!

PR #88, opened 2022-01-05
Fix `changeset` long-running PRs

joneshf-dd:
## Motivation

- Whenever a long-running PR branch was opened, we'd be skipping any of the `pull_request`/`push` workflows because we were making commits with `[skip ci]` and GitHub Actions automatically skip commits with that blessed text. We don't want that.

## Changes

- This changes things so we now make a commit with `[skip canary]` and check for that explicitly so we're not dealing with GitHub's blessed behavior.

## Testing

Nothing to check here.

## Releases

Choose one:

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -7,7 +7,7 @@ on:
 jobs:
     publish:
         runs-on: ubuntu-latest
-        if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
+        if: "!contains(github.event.head_commit.message, 'skip canary')"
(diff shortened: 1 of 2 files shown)


wfa207 (approved):
LGTM

joneshf-dd:
Thanks for the review!

Just noting that CI is broken at the moment as shown in https://github.com/DataDog/apps/runs/4721123363?check_suite_focus=true. Seems our packages not being consistent has finally bit us. We'll want to do https://datadoghq.atlassian.net/browse/UIAPPS-88 before going further. The only reason this PR didn't hit the same failure is that the commit message accidentally says `skip canary` so that job isn't running. Long story short, not merging this yet.

PR #98, opened 2022-01-22
UIAPPS-141 Detangle module dependencies

joneshf-dd:
## Motivation

- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-141
- We want to break the dependency cycles we have between most of the modules in the SDK.

## Changes

- We define a bunch of interfaces and use them to write code against rather than having everything depend on `DDClient`.
- We also enable the `import/no-cycle` rule so we don't backslide.

| Before                         | After                        |
| ------------------------------ | ---------------------------- |
| modules with a bunch of dependency cycles | modules without any depednecy cycles |

Legend:
- Green - no dependencies
- Purple - has normal dependencies
- Red - has circular dependencies

## Releases

Choose one:

<details>
  <br />

  ```bash
  npm install @datadog/create-app@1.0.3-canary.98.c0e1e09.0
  npm install @datadog/framepost@0.3.1-canary.98.c0e1e09.0
  npm install @datadog/ui-extensions-sdk@0.25.1-canary.98.c0e1e09.0
  # or 
  yarn add @datadog/create-app@1.0.3-canary.98.c0e1e09.0
  yarn add @datadog/framepost@0.3.1-canary.98.c0e1e09.0
  yarn add @datadog/ui-extensions-sdk@0.25.1-canary.98.c0e1e09.0
  ```
</details>

### DIFF
--- a/.changeset/twelve-ways-trade.md
+++ b/.changeset/twelve-ways-trade.md
@@ -2,6 +2,9 @@
 '@datadog/ui-extensions-sdk': minor
 ---
 
-Remove `RequireKeys` from exports.
-This is an internal helper.
-We don't actually want to export this and have it be part of the public API.
+-   Remove `RequireKeys` from exports.
+    This is an internal helper.
+    We don't actually want to export this and have it be part of the public API.
+-   Add `DebugClient` to break cyclical dependency.
+    This is not intended to be used outside of the SDK.
+    It is only intended to be used internally.
--- a/.changeset/twelve-ways-trade.md
+++ b/.changeset/twelve-ways-trade.md
@@ -5,7 +5,7 @@
 -   Remove `RequireKeys` from exports.
     This is an internal helper.
     We don't actually want to export this and have it be part of the public API.
--   Add `DebugClient`, `EventClient`, and `RequestClient` to break cyclical dependency.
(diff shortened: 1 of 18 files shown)


nmuldavin (commented):
I'm sorry for my late response on this one.

I have a (maybe) dumb question here: 

What's the difference between this change, in which we split the interface from the implementation, and what typescript does under the hood? To be honest i'm murky on what typescript actually does under the hood, but my working understanding is when you create a `class` it automatically creates both an interface and implementation, such that when you do `import type` you're actually importing the interface not the implementation.

joneshf-dd:
No worries. I honestly had no idea how TS was handling this. That's interesting to learn!

From what you describe, it doesn't seem like there's really much difference. The two things that really pop up for me are:
1. This is explicit and forces us to break the cycle ourselves. I think this helps with the mental model of how everything fits together.
2. We're a bit more granular since nothing requires the full public interface of a `DDClient` anymore. I think this opens the door to future simplifications/improvements/easier testing, but that remains to be seen.

nmuldavin (approved):
Well justified, let's see how it goes! If it feels like too much of a maintenance burden, i think it will be possible to re-tangle at least in the short term before there's any divergence between interface and implementations

joneshf-dd:
Thanks for the review!

PR #108, opened 2022-02-09
UIAPPS-147 Add `useCustomWidgetOptions` hook

joneshf-dd:
## Motivation

- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-147
- Much like the `useContext` hook did for `Context`, we want to abstract away some of the intricacies of dealing with custom widget options. It can be hard to get right, so we want a hook that makes it easier to do the right thing.

## Changes

- We add a new `useCustomWidgetOptions` hook that makes it easier to work with custom widget options.
- We also update the example Apps to use this new hook.

## Testing

- We added a suite of tests for this hook, so we should catch the majority of issues.

## Releases

Choose one:

<details>
  <br />

  ```bash
  npm install @datadog/create-app@1.1.1-canary.108.9412841.0
  npm install @datadog/ui-extensions-react@0.28.1-canary.108.9412841.0
  npm install @datadog/ui-extensions-sdk@0.28.1-canary.108.9412841.0
  # or 
  yarn add @datadog/create-app@1.1.1-canary.108.9412841.0
  yarn add @datadog/ui-extensions-react@0.28.1-canary.108.9412841.0
  yarn add @datadog/ui-extensions-sdk@0.28.1-canary.108.9412841.0
  ```
</details>


packages/create-app/src/template/package.json
```diff
@@ -3,8 +3,8 @@
     "version": "0.0.0",
     "private": true,
     "dependencies": {
-        "@datadog/ui-extensions-react": "0.26.0",
-        "@datadog/ui-extensions-sdk": "0.26.0",
+        "@datadog/ui-extensions-react": "0.28.0",
+        "@datadog/ui-extensions-sdk": "0.28.0",
```
joneshf-dd: **💭 thought:** ‏ These versions aren't quite right. They're what we predict the new version to be once this gets released. So like, pretty sure it'll be `0.28.0`, but not 100%. We should figure out a way to automatically update them like we do for the other versions in the repo.

joneshf-dd:
Actually, hold tight. I'm actually going to make a change to this API.

nmuldavin (approved):

joneshf-dd:
Thanks for the review!

We discussed this a bit on Slack/video, but I think a slightly different hook/set of hooks would be better here. Since we're going to hold off on changes for a bit, I'd rather we close this and revisit later. The rest of this comment explains the different hook/set of hooks.

---

Instead of needing to explicitly type things like:
```TypeScript
type Options = {
    option1: string;
    option2: boolean;
};

const options = useCustomWidgetOptions<Options>(client);
const option1: string | undefined = options.option1;
const option2: boolean | undefined = options.option2;
```

Grabbing individual options seems a bit more react-y (à la `useState`) and allow us to do a bit of parsing to make sure the options are the right types:

```TypeScript
const option1: string | undefined = useCustomWidgetOptionString(client, 'option1');
const option2: boolean | undefined = useCustomWidgetOptionBoolean(client, 'option2');
```

The main reason the current hook is questionable: we're effectively casting the options to whatever someone says the options are. With the different hooks, we could at the very least slap a `typeof options[name] === 'string' ? options[name] : undefined` or whatever in there to make sure that if we say it's a `string` it's actually a `string`. I feel like we can parlay those different hooks into a more robust design and maybe push it upstream to the SDK, but I don't want to get ahead of ourselves. In any case, something to worry about in the future.

PR #114, opened 2022-03-16
Deprecate dashboard_timeframe_change and dashboard_template_var_change events

nmuldavin:
## Motivation

Deprecate dashboard_timeframe_change and dashboard_template_var_change events in favor of the more consistent and general `context_change` event.

## Changes

Adds deprecation warnings on subscriptions to these events
Adds reporting of usage of deprecated events to web-ui with a new request type
Updates documentation, removing instructions related to these events

## Testing

Test app to come

## Releases

Choose one:

<details>
  <br />

  ```bash
  npm install @datadog/ui-extensions-react@0.29.1-canary.114.ea28f79.0
  npm install @datadog/ui-extensions-sdk@0.29.1-canary.114.ea28f79.0
  # or 
  yarn add @datadog/ui-extensions-react@0.29.1-canary.114.ea28f79.0
  yarn add @datadog/ui-extensions-sdk@0.29.1-canary.114.ea28f79.0
  ```
</details>

### DIFF
--- a/.changeset/fifty-spoons-drum.md
+++ b/.changeset/fifty-spoons-drum.md
@@ -1,0 +1,6 @@
+---
+'@datadog/ui-extensions-sdk': minor
+'@datadog/ui-extensions-react': minor
+---
+
+Deprecate granular dashboard change events in favor of context_change
--- a/docs/en/programming-model.md
+++ b/docs/en/programming-model.md
@@ -49,29 +49,33 @@ Although there are no architectural requirements for how to structure your app, 
 
 ## Context Data
 
-**Context** data provides information about the global and specific setting in which features mount in the Datadog UI. Context data will be provided to the SDK in several different ways:
+**Context** data provides information about the global and specific setting in which features mount in the Datadog UI. For example, in a custom widget the configured widget options will be present at `context.widget.options`, and on a dashboard the current template variables will be provided at `context.dashboard.templateVars`. 
 
-- When an App IFrame mounts and successfully handshakes with the Datadog UI, it will receive a set of **context** data containing basic information about the setting in which the IFrame renders. This context Data can be accessed with client.getContext() on the SDK client:
+### Getting initial context data
+
+When an App IFrame mounts and successfully handshakes with the Datadog UI, it will receive the initial **context**. This data can be accessed with client.getContext() on the SDK client:
 
 ```js
 client.getContext().then((context) => {
-  ...
+  console.log(context.dashboard.templateVars)
 })
 ```
 
-- When the context sent to an IFrame changes, each frame will receive new context from the `context_change` event:
+### Responding to changing context
+
+In most cases apps will also need to listen for changing context. For example, the current dashboard timeframe may change as the user edits. Whenever any subset of an iFrame's context changes, the Frame will receive a `context_change` event containing updated information. This can be used to update local state, respond to changes in specific fields, and more. 
 
 ```
 client.events.on('context_change', (newContext) => {
- //
+ if (newContext.dashboard.templateVars !== oldContext.dashboard.templateVars) {
+
+ }
 });
 ```
 
-- When menu-items (cog menu items, context menu items, or others) are clicked, the SDK in all active iframes will receive a click event. In order to determine where the click event is in the app, click event handlers will receive another set of context data:
+As a convenience for React developers, we provide a lightweight [useContext](https://datadoghq.dev/apps/modules/_datadog_ui_extensions_react.html#useContext) hook which automatically provides live-updated context data for use in a React component. 
 
-```js
-client.events.on('dashboard_cog_menu_click', (clickContext) => {});
-```
+If there is interest in similar framework-specific wrappers please let us know. 
 
 **Context structure attributes**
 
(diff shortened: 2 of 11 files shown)


docs/en/programming-model.md
```diff
@@ -49,29 +49,33 @@ Although there are no architectural requirements for how to structure your app,
 
 ## Context Data
 
-**Context** data provides information about the global and specific setting in which features mount in the Datadog UI. Context data will be provided to the SDK in several different ways:
+**Context** data provides information about the global and specific setting in which features mount in the Datadog UI. For example, in a custom widget the configured widget options will be present at `context.widget.options`, and on a dashboard the current template variables will be provided at `context.dashboard.templateVars`.
```
joneshf-dd: **❓ question:** ‏Is it worth calling out the combination of things? Like if you're using a custom widget on a dashboard you should have access to both `context.dashboard` and `context.widget`? Trying to step back into pre-Datadog Hardy's shoes, I don't think he would've assumed that you'd get both. Probably would've assumed that custom widgets just don't get access to `context.dashboard` no matter if they're on a dashboard or not.

docs/en/programming-model.md
```diff
@@ -49,29 +49,33 @@ Although there are no architectural requirements for how to structure your app,
...
 
-- When the context sent to an IFrame changes, each frame will receive new context from the `context_change` event:
+### Responding to changing context
+
+In most cases apps will also need to listen for changing context. For example, the current dashboard timeframe may change as the user edits. Whenever any subset of an iFrame's context changes, the Frame will receive a `context_change` event containing updated information. This can be used to update local state, respond to changes in specific fields, and more. 
 
 ```
 client.events.on('context_change', (newContext) => {
- //
+ if (newContext.dashboard.templateVars !== oldContext.dashboard.templateVars) {
```
joneshf-dd: **❓ question:** ‏If this were actually run, would it always be true? Since the `templateVars` is an array, wouldn't they always have different equality? I think for the sake of illustrating the idea this makes sense. I worry that someone (like myself) would grab this code, dump it into an App, and not understand why it wasn't working as intended.

I don't have any strong opinions about what to do instead. We could add a short comment noting that it's for illustrative purposes and to actually use a structural check (with like lodash) rather than an identity check. We could actually show it with a quick import of something like lodash. We could replace it with some call to a stubbed function. We could do something else.

.changeset/fifty-spoons-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+'@datadog/ui-extensions-sdk': minor
```
joneshf-dd, line 2: **❓ question:** ‏Can't comment on the actual lines. But since we're making a minor change anyway (and minor bumps in 0.x.x for SemVer are basically breaking changes), should we pull the `DASHBOARD_TIMEFRAME_CHANGE`, and `DASHBOARD_TEMPLATE_VAR_CHANGE` out of `EventType` and put them in some kind of `DeprecatedEventType` or something?

packages/ui-extensions-sdk/src/events/events.ts
```diff
@@ -49,6 +50,13 @@ interface DDEventDataTypes<AuthStateArgs> {
     [EventType.API_ACCESS_CHANGE]: APIAccessChangeEvent;
 }
 
+const deprecationWarnings: Partial<Record<EventType, string>> = {
```
joneshf-dd, line 53: **💭 thought:** ‏Unrelated to this change, but we should enable [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) at some point.

packages/ui-extensions-sdk/src/events/events.ts
```diff
@@ -49,6 +50,13 @@ interface DDEventDataTypes<AuthStateArgs> {
...
+    [EventType.DASHBOARD_TIMEFRAME_CHANGE]:
+        'The "dashboard_timeframe_change" event is deprecated. You may subcribe to updated dashboard timeframes from the more general "context_change" event.',
+    [EventType.DASHBOARD_TEMPLATE_VAR_CHANGE]:
+        'The "dashboard_template_var_change" event is deprecated. You may subscribe to template variable changes from the more general "context_change" event.'
```
joneshf-dd, line 57: **🔨 warning:** ‏I think this is going to create a bunch of warnings for the `useTemplateVariable` hook since it [listens to this event](https://github.com/DataDog/apps/blob/a4ebbdb28f09ae4ff7a41a3a5a630fc462fdc8c6/packages/ui-extensions-react/src/use-template-variable.ts#L75). We should be able to drop that event listener though and not worry about it, based on the motivation behind this change.

joneshf-dd (changes requested):
Looks good!

A bunch of ❓s that aren't blocking, so do whatever with them or merge without addressing. But there's one 🔨 that we should address before merging.

docs/en/programming-model.md
nmuldavin: You're right. I was trying to pick an example but that was a bad one

packages/ui-extensions-sdk/src/events/events.ts
nmuldavin: oh good catch. I'll update that hook to instead use `context`

.changeset/fifty-spoons-drum.md
nmuldavin: Let me think through what would happen here: 

For new apps, this would mean they get typescript errors if they try to use that event at all, which is the goal so that is good!

For existing apps on upgrade, this will mean they'll get typescript yelling at them so they will be FORCED to fix the problem. That's kinda nice, i suppose, but typescript won't have a [nice deprecation warning](https://github.com/DataDog/apps/pull/114/files#diff-33921cd9d0560e073ea9422e73bb99aa62be36d7ba833c14da81fb7afa698e4eR55), it will just yell at you but won't tell you how to fix it. That's not as nice of an experience. I'm torn between the forcing function and nice helptext. What do you think?

joneshf-dd: Ah, because TS will throw a TS2339 or something since the enum no longer has those properties? That's a good point. Maybe since we're still figuring out the deprecation story, we don't move things just yet. We can always do that later once we have a more clear picture.

packages/ui-extensions-react/src/use-template-variable.ts
```diff
@@ -70,29 +70,6 @@ function useTemplateVariable(
...
-
-                setVariable(value);
-            }
-        );
-
-        return () => {
-            unsubscribe();
-        };
-    }, [client.events, variableName]);
-
```
nmuldavin, line 95: @joneshf-dd I believe we are safe to _only_ use the context change event here, having verified that it always fires when template vars change

joneshf-dd: Yeah, that seems right! Thanks for grabbing this!

joneshf-dd (approved):
Looks good!

PR #117, opened 2022-03-18
Start on a go example

joneshf-dd:
## Motivation

- We want to show that it's possible to build an App without writing a ton of JavaScript.

## Changes

- We create a very simple App in go that only uses the `"dashboard_cog_menu"` feature. This App has a single item that when clicked will search for a specific template variable, then send a notification back to Datadog with the value of that template variable. This shows that it's entirely possible to communicate back and forth between golang and JavaScript.
- The magic that makes this happen is that go can compile to WASM easily. We take the WASM, run it in the browser, and the rest is history.
- This is not really usable as it is. If we want to get serious about this, we should extract a go package that actually makes it easier to use this stuff. Likely we'd be able to get away with just doing a little JSON serialization/deserialization to plain go types. This mainly exists as a proof-of-concept.

example of the App running

## Testing

- TODO

## Releases

Choose one:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -34,6 +34,10 @@
     "editor.codeActionsOnSave": {
         "source.fixAll.eslint": true
     },
+    "go.toolsEnvVars": {
+        "GOARCH": "wasm",
+        "GOOS": "js"
+    },
     "prettier.prettierPath": ".yarn/sdks/prettier/index.js",
     "prettier.configPath": "prettier.config.js",
     "search.exclude": {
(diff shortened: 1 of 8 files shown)


joneshf-dd:
* master
  * **PR #117** <a href="https://app.graphite.dev/github/pr/DataDog/apps/117" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/apps/117?utm_source=stack-comment).

joneshf-dd:
Not working on this currently, so closing this PR.

PR #115, opened 2022-03-17
Make an example with the TypeScript API client

joneshf-dd:
## Motivation

- The `@datadog/datadog-api-client` is to the point where it's possible to use it in a browser now!

## Changes

- We're adding an example of integrating with the `@datadog/datadog-api-client`.

## Testing

- TODO

## Releases

Choose one:

-   [ ] No release is necessary.
        If you're only updating examples/documentation, this is likely what you want.
-   [x] All packages that need a release have a changeset.

PR #120, opened 2022-03-24
UIAPPS-165 Start custom widget with 1st-party data

joneshf-dd:
Motivation
----------

- We want to make a template for a custom widget that interacts with 1st-party data. This should continue to grow and change as features evolve.
- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-165

Changes
-------

- We create an example we can use as a template that solves a specific use case: an App where there's a custom widget that interacts with 1st-party data.
- We are hitting the monitors read API endpoint because it's a relatively straight-forward endpoint. We also try to explain the OAuth client and scope necessary.
- We make a few changes from the typical examples we have. An inexhaustive list of some of the changes (in no particular order):

    - We centralize the client initialization logic. This is something that the stream example does and it seems to make for a better overall experience. The type inference is better since the auth logic can actually infer the type to be used elsewhere in the App.
    - We don't do dynamic imports. This is another thing the stream example does and it again seems to be more conventional. I'm sure there's a reason the other examples do dynamic imports, but it's not clear to me what that reason is. If there's a hard requirement why we need that, we can of course move to that.
    - We split up more of the logic across files. One of the things that's kind of hard about the current examples (at least to me) is that they tend to have a lot of behavior in a few files. We try to split along responsibilities, though the boundaries might not be 100% correct.
    - We handle more of the unhappy path. Most of the examples don't seem to actually deal with edge cases/error cases too robustly.  Since we expect people to actually turn this template into a real App, we want to start them off on the right foot by at least showing where the edge cases/error cases live.
    - We explicitly setup formatting/linting as `npm scripts`. Most of the other examples rely on the fact that they exist in this monorepo and so their formatting/linting will happen automatically. Since we expect people to build something outside of this monorepo, we want to start them off on the right foot with some scripts that can help.
- We don't use the official Datadog API client to interact with the 1st-party data. The reason is largely because the story is still being written in https://github.com/DataDog/apps/pull/115. Once we have something that works a little better, we should switch this template over to using it. But for now, it's better to at least have _something_ than stymie on getting this template done.

Testing
-------

Releases
--------

Choose one:

<details>
  <br />

  ```bash
  npm install @datadog/create-app@1.1.1-canary.120.07b6648.0
  npm install @datadog/ui-extensions-react@0.30.1-canary.120.07b6648.0
  # or 
  yarn add @datadog/create-app@1.1.1-canary.120.07b6648.0
  yarn add @datadog/ui-extensions-react@0.30.1-canary.120.07b6648.0
  ```
</details>


joneshf-dd:
* master
  * **PR #120** <a href="https://app.graphite.dev/github/pr/DataDog/apps/120" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://graphite.dev?utm_source=gh).

examples/custom-widget-with-1st-party-data/src/controller/controller.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    return (
+        <>
+            <div>The application controller is running in the background.</div>
+            <a href="http://localhost:3000/custom-widget">
+                Click here to open your custom widget
+            </a>
+        </>
+    );
+}
```
nmuldavin: I love this. Even though the controller doesn't need to be react, making it react anyways for consistency is great

examples/custom-widget-with-1st-party-data/src/controller/custom-widget.ts
```diff
@@ -0,0 +1,7 @@
+import { DDClient } from '@datadog/ui-extensions-sdk';
+
+/**
+ * Here we do any initialization of we need for the custom widget.
+ * This could include listening to events from the main controller
+ */
+export function setupCustomWidget(client: DDClient): void {}
```
nmuldavin: Do you have specific ideas what is needed here? We could remove it until it's needed

I have one idea but we can't do it until we complete the widget options work you started earlier this querter: Dyamic widget options

nmuldavin (approved):
This rocks!

examples/custom-widget-with-1st-party-data/src/controller/custom-widget.ts
joneshf-dd: Yeah, that was the main thing I was thinking. But yeah, probably not worth hanging onto this.

Would the `api_access_change` event be worthwhile to illustrate here? It is about 1st-party auth (as far as I understand).

joneshf-dd: Whether we keep this or not, I realize it might actually be better to have these setup functions as `useSetup*` hooks that actually do the right thing as far as unsubscribing of listeners.

joneshf-dd: Eh, let's just dump it. The other templates can show this pattern, and if we ever move on UIAPPS-90, we can follow that pattern.

joneshf-dd:
Thanks for the review!

PR #121, opened 2022-03-24
UIAPPS-147 Add hooks for custom widget options

joneshf-dd:
Motivation
----------

- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-147
- Previous PR: https://github.com/DataDog/apps/pull/108
- Much like the `useTemplateVariable` hook did for working with template variables, we want to abstract away some of the intricacies of dealing with custom widget options. It can be hard to get right, so we want a hook that makes it easier to do the right thing.

Changes
-------

- We add three new hooks that make it easier to work with custom widget options.
    - The `useCustomWidgetOption` hook is the main implementation. It supports a parser that is used by the others while also making it possible to refine values a bit better.
    - The `useCustomWidgetOptionBoolean` and `useCustomWidgetOptionString` hooks provide easier to use helpers for the common case of dealing with a boolean or string options, respectively.
- These hooks implement the updated API as described in https://github.com/DataDog/apps/pull/108#issuecomment-1042038199.
- We also update the example Apps to use these new hooks.

Testing
-------

- We added a suite of tests for these hooks, so we should catch the majority of issues.

Releases
--------

Choose one:

-   [ ] No release is necessary.
        If you're only updating examples/documentation, this is likely what you want.
-   [x] All packages that need a release have a changeset.

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @datadog/create-app@1.1.1-canary.121.b4c4373.0
  npm install @datadog/ui-extensions-react@0.30.1-canary.121.b4c4373.0
  # or 
  yarn add @datadog/create-app@1.1.1-canary.121.b4c4373.0
  yarn add @datadog/ui-extensions-react@0.30.1-canary.121.b4c4373.0
  ```
</details>

PR #123, opened 2022-03-24
Remove some dead code from Geomap example

joneshf-dd:
Motivation
----------

- While updating https://github.com/DataDog/apps/pull/121, we found some dead code.

Changes
-------

- We remove what looks like a bunch of dead code in this example. The check for whether it's dead is what goes on in `examples/geomap/setup/createDatadogApp.js`. Since the only feature present is the single Geomap widget, everything else seems dead.
- If some of this is actually necessary, we can add it back.

Testing
-------

- Probably defer to Tom for any testing we want to perform here.

Releases
--------

Choose one:


joneshf-dd:
Requesting a review from Tom, since this is a change to the Geomap example.

tdimnet (approved):
LGTM 👍

joneshf-dd:
Thanks for the review!

PR #124, opened 2022-03-25
Security network resource monitoring

nmuldavin:
Network resource monitoring

this PR adds monitoring of all network resources loaded via the performance api. Collection is done in batches and reported back to web-ui for logging every 10 seconds with an event.

<details>
  <br />

  ```bash
  npm install @datadog/ui-extensions-sdk@0.31.1-canary.124.2dcecab.0
  # or 
  yarn add @datadog/ui-extensions-sdk@0.31.1-canary.124.2dcecab.0
  ```
</details>

### DIFF
--- a/packages/ui-extensions-sdk/src/types.ts
+++ b/packages/ui-extensions-sdk/src/types.ts
@@ -343,7 +343,8 @@ export interface DeprecatedEventUsage {
 export type DeprecatedUsage = DeprecatedEventUsage; // we can union type here later
 
 export interface LoadedResourceMetadata {
-    url: string; // the resources URL
+    url?: string; // the resources URL (set to null in case of network calls for privacy reason)
+    urlHostname: string; // the resource URL hostname
     initiatorType: string; // the type of resource that initiated the performance event (xmlhttprequest, css, img)
     nextHopProtocol: string; // the network protocol used to fetch the resource, as identified by the ALPN Protocol ID (RFC7301).
 }
--- a/packages/ui-extensions-sdk/src/types.ts
+++ b/packages/ui-extensions-sdk/src/types.ts
@@ -343,11 +343,14 @@ export interface DeprecatedEventUsage {
 export type DeprecatedUsage = DeprecatedEventUsage; // we can union type here later
 
 export interface LoadedResourceMetadata {
-    timestamp: number;
+    startTimeTs: number; // timestamp (in ms) of the resource started being fetched
+    secureConnectionStartTs: number; // timestamp (in ms) of the SSL handshake
     url?: string; // the resources URL (set to null in case of network calls for privacy reason)
     urlHostname: string; // the resource URL hostname
     initiatorType: string; // the type of resource that initiated the performance event (xmlhttprequest, css, img)
     nextHopProtocol: string; // the network protocol used to fetch the resource, as identified by the ALPN Protocol ID (RFC7301).
+    duration: number; // overall time required to fetch the resource
+    decodedBodySize: number; // size of the body after removing any applied content-codings
 }
 
 export interface LoadedResourceMetaDataBatch {
--- a/packages/ui-extensions-sdk/src/types.ts
+++ b/packages/ui-extensions-sdk/src/types.ts
@@ -345,7 +345,7 @@ export type DeprecatedUsage = DeprecatedEventUsage; // we can union type here la
 export interface LoadedResourceMetadata {
     startTimeTs: number; // timestamp (in ms) of the resource started being fetched
     secureConnectionStartTs: number; // timestamp (in ms) of the SSL handshake
-    url?: string; // the resources URL (set to null in case of network calls for privacy reason)
+    url?: string; // the resources URL (set to null in case of network calls for privacy reason). This value doesn't change even if the request is redirected.
     urlHostname: string; // the resource URL hostname
     initiatorType: string; // the type of resource that initiated the performance event (xmlhttprequest, css, img)
     nextHopProtocol: string; // the network protocol used to fetch the resource, as identified by the ALPN Protocol ID (RFC7301).
--- a/packages/ui-extensions-sdk/src/types.ts
+++ b/packages/ui-extensions-sdk/src/types.ts
@@ -356,5 +356,3 @@ export interface LoadedResourceMetadata {
 export interface LoadedResourceMetaDataBatch {
     resources: LoadedResourceMetadata[];
 }
-
-export interface NetworkRequestMetadata {}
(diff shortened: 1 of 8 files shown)


joneshf-dd (approved):
Looks good!

packages/ui-extensions-sdk/src/types.ts
```diff
@@ -341,3 +341,18 @@ export interface DeprecatedEventUsage {
...
+
+export interface LoadedResourceMetadata {
+    startTimeTs: number; // timestamp (in ms) of the resource started being fetched
+    secureConnectionStartTs: number; // timestamp (in ms) of the SSL handshake
+    url?: string; // the resources URL (set to null in case of network calls for privacy reason). This value doesn't change even if the request is redirected.
+    urlHostname: string; // the resource URL hostname
+    initiatorType: string; // the type of resource that initiated the performance event (xmlhttprequest, css, img)
+    nextHopProtocol: string; // the network protocol used to fetch the resource, as identified by the ALPN Protocol ID (RFC7301).
+    duration: number; // overall time required to fetch the resource
+    decodedBodySize: number; // size of the body after removing any applied content-codings
```
joneshf-dd, line 353: **💭 thought:** ‏If these comments are turned into JSDoc format, they will show up in editors (like VS Code). Not blocking, so up to you whether you want to change it or not.

packages/ui-extensions-sdk/src/utils/security.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        initiatorType: 'fetch',
+        name: 'http://localhost/api/1'
+    }
+};
+
+beforeEach(() => {
+    performance.getEntriesByType.mockReset();
+});
+
+describe('collectResourceUsage', () => {
```
joneshf-dd: **👏 praise:** ‏These tests are great!

joneshf-dd (approved):
This looks good!

I gotta be honest with you, a lot of the implementation went over my head for lack of familiarity with the Performance API. Given that multiple eyes have been on this, it seems fine to me. But if you'd like a more in-depth review, I can read up on the API and give this another go.

packages/ui-extensions-sdk/src/utils/security.ts
```diff
@@ -0,0 +1,113 @@
...
+    const snapshot = performance.getEntriesByType('resource');
+
+    const resources = snapshot
+        .filter(isResourceTimingEntry)
+        // to avoid collecting the same resource twice, we use
+        // the Set difference: snapshot - previousSnapshot
+        .filter(r => !previouslyLoadedResources.has(r.name))
+        .map(r => {
+            // safe because r.name represents the resolved URL of the requested resource.
+            const urlHostname = new URL(r.name).hostname;
```
bcaudan: **❓ question:** ‏what browsers are you targeting? (`URL` is not available in IE)

packages/ui-extensions-sdk/src/utils/security.ts
```diff
@@ -0,0 +1,113 @@
...
+        loadedResourceIds = newIds;
+
+        // send to web-ui
+        reportUsage(batch);
+    };
+
+    const interval = setImmediateInterval(() => {
+        try {
+            doResourceCollection();
+        } catch (e) {
```
bcaudan: **💬 suggestion:** ‏You may want to consider [PerformanceObserver](https://developer.mozilla.org/fr/docs/Web/API/PerformanceObserver)

packages/ui-extensions-sdk/src/utils/security.ts
```diff
@@ -0,0 +1,113 @@
...
+            }
+            return {
+                startTimeTs: ts + r.startTime,
+                secureConnectionStartTs: ts + r.secureConnectionStart,
+                url,
+                urlHostname,
+                initiatorType: r.initiatorType,
+                nextHopProtocol: r.nextHopProtocol,
+                duration: r.duration,
+                decodedBodySize: r.decodedBodySize
```
bcaudan: **❗ warning:** ‏FYI, most of the timing / size attributes won't be available for [cross origin resources](https://www.w3.org/TR/resource-timing-2/#sec-cross-origin-resources)

packages/ui-extensions-sdk/src/utils/security.ts
adrienperonnetDT: That looks great and will remove our polling/deduplication logic, @nmuldavin I've tested it on my branch : https://github.com/DataDog/apps/pull/143, feel free to merge onto yours.

packages/ui-extensions-sdk/src/utils/security.ts
nmuldavin: No IE 🙏

packages/ui-extensions-sdk/src/utils/security.ts
nmuldavin: Very cool, good thing you consulted the expert :) I'll review your PR now

packages/ui-extensions-sdk/src/utils/security.ts
nmuldavin: Thanks, that should be fine, it just means incomplete data in some cases

kathy-dd (approved):

packages/ui-extensions-sdk/src/utils/security.ts
```diff
@@ -42,44 +42,36 @@ export const parseResource = (
...
 
     const observer = new PerformanceObserver(observerCallback);
     observer.observe({ type: 'resource', buffered: true });
 
-    let bufferFullHandler = () => {};
-
-    if ('addEventListener' in performance) {
-        bufferFullHandler = () => {
-            performance.clearResourceTimings();
-        };
```
adrienperonnetDT: @nmuldavin It's not clear to me how the `PerformanceObserver` will behave in case resource timing buffer is full. Reading at https://chromestatus.com/feature/5320666234486784 my understanding is that we could miss some entries if the buffer is full. If this is the case I wonder if we should keep the call to `clearResourceTimings` to handle that case ?

adrienperonnetDT: Resolving my comment, from the [spec](https://w3c.github.io/performance-timeline) it seems Performance observer buffer is emptied after the observers have been notified, so no need to clear it.

adrienperonnetDT (approved):

packages/ui-extensions-sdk/src/utils/security.ts
nmuldavin: yeah i did some reading on this, i think we are ok. 

There is one possible edge-case, which is that the buffer fills before the observer is instantiated such that the initial call to `performance.getEntriesByType()` isn't complete, but this is very unlikely given that the SDK must be instantiated early

PR #125, opened 2022-03-30
UIAPPS-164 Start custom widget with 3rd-party data

joneshf-dd:
Motivation
----------

- We want to make a template for a custom widget that interacts with 3rd-party data. This should continue to grow and change as features evolve.
- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-164

Changes
-------

- We create an example we can use as a template that solves a specific use case: an App where there's a custom widget that interacts with 3rd-party data.
- We're following a lot of the setup from the custom widget with 1st-party data from https://github.com/DataDog/apps/pull/120.

Testing
-------

Releases
--------

Choose one:


joneshf-dd:
* master
  * **PR #125** <a href="https://app.graphite.dev/github/pr/DataDog/apps/125" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/apps/125?utm_source=stack-comment).

examples/custom-widget-with-3rd-party-data/src/client/client.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * We initialize the {@link DDClient} in one place and use it throughout the App.
+ * Having the {@link DDClient} initialized in one place helps centralize auth logic and provide better type inference.
+ */
+const client: DDClient<User> = init({
+    authProvider: {
+        /**
+         * The `authStateCallback` can be pretty simple,
+         * but can do whatever you need to authenticate.
```
nmuldavin: I think we should explain that this is to check the authentication status, rather than to do the authentication itself

examples/custom-widget-with-3rd-party-data/src/client/client.ts
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+            return {
+                args: user,
+                isAuthenticated: true
+            };
+        },
+        /**
+         * We use `'close'` so the login page can close the window to notify successful login.
+         * Once that happens, the `authStateCallback` is invoked again to check the state.
```
nmuldavin, line 26: Maybe provide a link to the relevant docs section for additional reading: https://github.com/DataDog/apps/blob/master/docs/en/programming-model.md#authentication

examples/custom-widget-with-3rd-party-data/src/controller/custom-widget.ts
```diff
@@ -0,0 +1,47 @@
...
+                    return;
+                }
+
+                /**
+                 * Perform the actual logout,
+                 * then make sure to update the auth state so it's reflected in Datadog.
+                 */
+                await logoutUser();
+                await client.auth.updateAuthState();
+            }
```
nmuldavin, line 35: I love it, it's never been clear how best to handle this. Devon suggested making 'logout' a first-class thing defined in the auth provider, but until then having this example is great

examples/custom-widget-with-3rd-party-data/src/custom-widget/custom-widget.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * If we do not have a {@link User},
+     * then authentication was not successful.
+     * Datadog will not actually render the custom widget in this case,
+     * but we still have to handle this case.
+     */
+    if (user == null) {
+        return <></>;
+    }
```
nmuldavin: This is interesting. I feel like the whole point of our auth system is that the app already has information about the user, and just wants to let us know about it so we can display a connect button or not. Is it the right pattern to get that data from the auth state args instead of directly from localstorage?

I have to be honest that the utility of auth state args and also `getAuthState()` has always felt murky to me, we sorta did it from a 'why not?' point of view and to be consistent with the args pattern for sidepanels + modals

examples/custom-widget-with-3rd-party-data/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    case '/custom-widget':
+        renderCustomWidget(client);
+        break;
+
+    /**
+     * This route is not really part of the App.
+     * It's only provided for completeness.
+     * In an actual App,
+     * this should be removed and the actual authentication route should be used.
+     */
```
nmuldavin: Login *could* be part of the app if they wanted it to be. Maybe the right wording is about that option? So something like 'Here the login route is part of the datadog app but the system is designed to be able to leverage  preexisting login route if you have access to it'

nmuldavin (commented):
This is 🔥 I left some thoughts, let me know what you think!

examples/custom-widget-with-3rd-party-data/src/custom-widget/custom-widget.tsx
joneshf-dd: This is good background! As someone that wasn't around for these discussions, I assumed we wanted App developers to store information in `AuthState` so they can use it in a different place in the App. Interesting to know that was not a hard requirement. I don't personally have a preference one way or the other. I think some things are harder one way and easier the other way.

However, I think if we don't expect or want App developers to store things in `AuthState`, we should drop it in exchange for something else (likely a `boolean`), get rid of the rest of the stuff like `DDAuthClient.getAuthState`/`DDAuthClient.updateAuthState`, and reword the methods to focus on auth rather than the state of the auth (e.g. rename `authStateCallback` to `isAuthenticated`). If all the App is able to do is say whether the user is authenticated or not, it seems hard to justify methods for other things around the auth state, or even hint at the idea that an App might be visible if it requires 3rd-party auth, but isn't actually authenticated.

For the sake of moving this template forward, I'm going to change this stuff to not use the `AuthState`.

joneshf-dd:
Mind giving it another look?

examples/custom-widget-with-3rd-party-data/src/custom-widget/custom-widget.tsx
nmuldavin: I'm open to removing those hooks + auth args, I'll make a ticket

nmuldavin (approved):
This is great, thanks for the changes

joneshf-dd:
Thanks for the review!

PR #126, opened 2022-03-30
chore: update react-scripts to v5

ericrallen:
## Motivation

This updates react-scripts in the generated starter app from `@datadog/create-app` in order to resolve the `react-error-overlay` issue that was preventing hot reloading from working while developing an app after using `yarn create @datadog/app`.

**References**:
- https://github.com/facebook/create-react-app/issues/11771
- https://stackoverflow.com/a/70452191/656011

## Changes

Upgrade to `v5.0.0` of `react-scripts`

### To See the Current Issue

1. Clear yarn cache and run `yarn create @datadog/app`
2. Run `yarn start` in generated app
3. Make change to `src/widget/index.tsx` in generated app
4. Notice that hot reload does not work

### To Test the Idea Behind This Fix

1. Delete `node_modules` and `yarn.lock`
2. Update `react-scripts` version in `package.json` to `^5.0.0`
3. Run `yarn`
4. Repeat steps 2 & 3 from the [To See the Current Issue](#to-see-the-current-issue) above
5. Notice that hot reload does work

**To Test this Fix**:

1. Checkout this branch
2. `yarn link` the `@datadog/create-app` package
3. Run `yarn create @datadog/app --commit 52f5e05d00077915e8a7ed424d4bd83ae6e1fe85`
4. Run `yarn start` in generated app
5. Make a change to `src/widget/index.tsx` in generated app
6. Notice that hot reload does work
7. Shut down the development server
8. `yarn unlink` the `@datadog/create-app` package

## Releases

Choose one:

-   [x] All packages that need a release have a changeset
        Only  `@datadog/create-app` needs a release


ericrallen:
Here's s screenshot of the previous behavior where the `react-error-overlay` `<iframe>` was blocking the DOM and preventing user interaction, too:

Screen Shot 2022-03-30 at 10 24 45

packages/create-app/src/template/package.json
```diff
@@ -11,7 +11,7 @@
         "milligram": "^1.4.1",
         "react": "^17.0.1",
         "react-dom": "^17.0.1",
-        "react-scripts": "4.0.3",
+        "react-scripts": "^5.0.0",
```
joneshf-dd: Unfortunately, I think the actual change is going to need to be a bit bigger. There's two things that the `@datadog/create-app` package does: create an App from an example in the `examples` directory, piece together an App from the template in the `packages/create-app/src/template` directory. The change here only affects the latter use case. Given that your reproduction steps are to use `yarn create @datadog/app`, it looks like you're wanting to do the former use case (which defaults to the `examples/starter-kit` example). In order to address the `yarn create @datadog/app` use case, we'd actually want to update `examples/starter-kit/package.json` in addition to this change.

---

Now the part where it becomes a bigger change than expected. Since we're using yarn classic, and we're using workspaces, and we're using `react-scripts` in most (all?) of the examples, we run into some issues with versions. `react-scripts` wants all of its dependencies to match exactly. If we try to update one example's version of `react-scripts`, it will pull in a different version of dependencies (like `webpack`) and then it will fail with a pretty cryptic error message. My understanding is that yarn classic sometimes hoists dependencies to the root, and `react-scripts` throws an error when there are two different versions somewhere in the dependency chain–even if it _shouldn't_ be able to see it due to them being in different `node_modules` directories.

This problem could probably be solved in a lot of different ways: we could update to yarn berry, but then we run into other dependencies breaking because they don't support yarn berry; we could investigate further about how to prevent hoisting; we could not use `react-scripts` so heavily; we could likely do something else.

The best idea we've come up for this is to update all of the `react-scripts` to the same version. Relatedly, this is also why we're specifying an exact version for `react-scripts`. We didn't do that in the past and when we tried to setup workspaces it all went sideways because one example was using `4.0.x` and another using `4.0.y`. So for the sake of not causing similar issues down the road, we'll want to make sure we use an exact version of the `5.x.y` series.

The reason we haven't done this already is that we haven't looked into how much work it is to do. It might be a one-line change in each of the examples that depend on `react-scripts`. Or, there might be some transitive chain of dependency issues we have to figure out.

---

As far as what to do going forward. It seems like we should update all of the `react-scripts` dependencies to be some exact version of the `5.x.y` series. If this sounds like too much work (since it's unknown what the fallout is) feel free to push back on us to figure it out. I'm more than happy to do this work, but don't want to start on it without checking with you first. If this seems like something you want to do, have at it. It could be really simple! Just don't want you to feel stuck with this ball of yarn because you pulled the thread.

joneshf-dd: It looks like the latest version is `5.0.0`.

```suggestion
        "react-scripts": "5.0.0",
```

joneshf-dd (commented):
Sorry this broke on you. But, thanks for looking into it!

There's a bit more work before we can merge it. I tried to explain the situation and some ideas of what we can do, totally up to you what you'd like here.

packages/create-app/src/template/package.json
ericrallen: Thanks for all the context.

I was a little worried that it was going to be a slightly larger issue due to the workspace.

I'll take a swing at getting everything updated and make sure it all still works as expected.

packages/ui-extensions-react/package.json
```diff
@@ -38,7 +38,7 @@
     },
     "peerDependencies": {
         "@datadog/ui-extensions-sdk": "0.30.1",
-        "react": "^17.0.2"
+        "react": "^18.0.2"
```
joneshf-dd: Since we're updating the peer dependencies, we're going to want to give this package a minor bump (minor only because we're still `0.x.y` on the `@datadog/ui-extensions-*` packages). And since this package is getting a bump, we'll also need to bump `@datadog/ui-extensions-sdk`. CI won't pass unless they're both bumped, so it's not really something you should be able to mess up, more a heads up.

Not sure what your familiarity is with `changesets`, but you can either add another changeset with `yarn changeset add` selecting these two packages or edit the front matter in the changeset in this PR: `.changeset/cuddly-pugs-hammer.md`. I'll make a [suggestion for the latter](https://github.com/DataDog/apps/pull/126#discussion_r841994846), but feel free to ignore it and do the former if you prefer.

Also if you'd prefer to avoid this whole thing, just merge it now and we'll follow-up on the version bumps before making the next release.

.changeset/cuddly-pugs-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@datadog/create-app': major
```
joneshf-dd, line 2: Bumping the other packages due to peer dependencies updates.

```suggestion
'@datadog/create-app': major
'@datadog/ui-extensions-react': minor
'@datadog/ui-extensions-sdk': minor
```

Again, feel free to ignore this, or make a separate changeset, if you'd prefer.

joneshf-dd (approved):
This looks great! Thanks for going through this.

One last bit about bumping the packages. But as mentioned, if you'd prefer to avoid making more changes/dealing with versions, just merge it now and we'll follow-up on the version bumps before making the next release.

.changeset/cuddly-pugs-hammer.md
ericrallen: @joneshf-dd Good catch! After going down the dependency rabbit hole I totally forgot to add a changeset for the other packages I had to touch.

I'll get that updated and pushed out.

joneshf-dd (approved):
Thanks again for doing all this!

PR #44, opened 2021-08-10
Remove dynamic widgets

enbashi:
Remove the ability to define custom widgets by code: `client.dashboard.customWidget.onRequest`.

PR #128, opened 2022-04-04
implement collectResourceUsage

adrienperonnetDT:
## Motivation

-   What is the motivation / context for this change? (Text, screenshot, JIRA ticket, Github Issue)

## Changes

-   What does this change? (Link to Github Issue if applicable, e.g. #44)

| Before                         | After                        |
| ------------------------------ | ---------------------------- |
| Before alt text | After alt text |

## Releases

Choose one:

<details>
  <br />

  ```bash
  npm install @datadog/ui-extensions-sdk@0.30.2-canary.128.466946c.0
  # or 
  yarn add @datadog/ui-extensions-sdk@0.30.2-canary.128.466946c.0
  ```
</details>

### DIFF
--- a/packages/ui-extensions-sdk/src/client/client.ts
+++ b/packages/ui-extensions-sdk/src/client/client.ts
@@ -220,9 +220,6 @@ export class DDClient<AuthStateArgs = unknown>
         let loadedResourceIds: LoadedResourceIds = new Set();
 
         if (!this.supportPerformanceObject()) {
-            this.logger.warn(
-                'resource monitoring is disabled because browser does not support performance object'
-            );
             return;
         }
 
--- a/packages/ui-extensions-sdk/src/client/client.ts
+++ b/packages/ui-extensions-sdk/src/client/client.ts
@@ -212,34 +212,37 @@ export class DDClient<AuthStateArgs = unknown>
         this.debug = context?.app?.debug || this.debug;
     }
 
+    private supportPerformanceObject() {
+        return window.performance !== undefined && 'getEntries' in performance;
+    }
+
     private startResourceMonitoring() {
         let loadedResourceIds: LoadedResourceIds = new Set();
 
-        if (
-            !(
-                window.performance !== undefined &&
-                'getEntriesByType' in performance
-            )
-        ) {
+        if (!this.supportPerformanceObject()) {
             this.logger.warn(
                 'resource monitoring is disabled because browser does not support performance object'
             );
             return;
         }
 
+        const collectResources = async () => {
+            // collect batch of resource-loading data
+            const [batch, newIds] = collectResourceUsage(loadedResourceIds);
+
+            // update index of loaded resources
+            loadedResourceIds = newIds;
+
+            // send to web-ui
+            await this.framePostClient.request(
+                RequestType.SECURITY_LOG_RESOURCES_LOADED,
+                batch
+            );
+        };
+
         const interval = setImmediateInterval(async () => {
             try {
-                // collect batch of resource-loading data
-                const [batch, newIds] = collectResourceUsage(loadedResourceIds);
-
-                // update index of loaded resources
-                loadedResourceIds = newIds;
-
-                // send to web-ui
-                await this.framePostClient.request(
-                    RequestType.SECURITY_LOG_RESOURCES_LOADED,
-                    batch
-                );
+                await collectResources();
             } catch (e) {
                 // Stop batch collecting if there's an error
                 clearInterval(interval);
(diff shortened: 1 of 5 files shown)


packages/ui-extensions-sdk/src/utils/security.ts
```diff
@@ -1,19 +1,28 @@
...
-    const newLoadedResourceIds = new Set(previouslyLoadedResources);
+    // Use resource timing api to to retrieve application loaded resources
 
-    // TODO: Do batch collection of resource data
-    // TODO: Update list of preivously loaded resource ids
+    const snapshot = performance.getEntriesByType('resource').map(e => {
+        const a = e as PerformanceResourceTiming;
+        return {
+            url: a.name,
+            initiatorType: a.initiatorType,
```
kathy-dd: If the initiator type is `FETCH` let's remove the path from the URL for privacy reasons.  Not sure where you can inject that logic?  Maybe in the interface?  Should we consider any other `initiatorType`?

adrienperonnetDT: `xmlhttprequest` and `fetch` seems to be the only two request-related initiator types.
Also RUM is doing similar filtering in : 
https://github.com/DataDog/browser-sdk/blob/cb1918e8a94a9fdabef9fcc3396a2bd0ddf8ad62/packages/rum-core/src/domain/rumEventsCollection/resource/resourceUtils.ts#L74-L76

packages/ui-extensions-sdk/src/client/client.ts
```diff
@@ -212,27 +212,53 @@ export class DDClient<AuthStateArgs = unknown>
...
         let loadedResourceIds: LoadedResourceIds = new Set();
 
-        const interval = setImmediateInterval(async () => {
-            try {
-                // collect batch of resource-loading data
-                const [batch, newIds] = collectResourceUsage(loadedResourceIds);
+        if (!this.supportPerformanceObject()) {
+            this.logger.warn(
+                'resource monitoring is disabled because browser does not support performance object'
+            );
```
nmuldavin: I'm not sure we should log here: logs in the SDK are for developers, and i'm not sure they care very much whether resource monitoring is active or not.

packages/ui-extensions-sdk/src/client/client.ts
```diff
@@ -212,27 +212,53 @@ export class DDClient<AuthStateArgs = unknown>
...
+        if ('addEventListener' in performance) {
+            performance.addEventListener(
+                'resourcetimingbufferfull',
+                async () => {
+                    // ensure we collect the last resources before clearing the buffer
+                    await collectResources();
+                    performance.clearResourceTimings();
+                }
+            );
+        }
```
nmuldavin, line 253: Makes sense!

packages/ui-extensions-sdk/src/utils/security.ts
```diff
@@ -1,19 +1,41 @@
...
+            return {
+                url,
+                urlHostname,
+                initiatorType: a.initiatorType,
+                nextHopProtocol: a.nextHopProtocol
+            };
+        });
 
-    return [{ resources: [] }, newLoadedResourceIds];
+    return [{ resources }, new Set(snapshot.map(x => x.name))];
```
nmuldavin: concern: You are setting the 'loadedResourceIds' to be the names of the current `snapshot`, but don't you want to accumulate _all_ of the ids you've ever loaded?

```
const updatedIds = new Set(previouslyLoadedResources);

snapshot.forEach(item => { updatedIds.add(item.name) })
```

adrienperonnetDT: You're right I'm setting the `loadedResourceIds` from the current snapshot instead of accumulating all previously loaded resources.

I think deduplication would still work with that approach because
- before `resourcetimingbufferfull` triggers : `previouslyLoadedResources` is always subset of `performance.getEntriesByType('resource')`, so setting the `loadedResourceIds` to current snapshot should be exactly the same as accumulating all previous IDs.
- If `resourcetimingbufferfull` triggers : performances entries buffer will be reset and `collectResources` will not report resources that have already been reported so we have no risk of sending duplicate resources.

This approach will avoid having an unbounded `loadedResourceIds` Set, it will be reseted when `resourcetimingbufferfull` is reached.

But maybe this is misleading/obscure given the name `loadedResourceIds` which might indicates it contains all previous ids. Do you think renaming var `loadedResourceIds` to something like `previousSnapshotResourceIds` would help ?

packages/ui-extensions-sdk/src/client/client.ts
adrienperonnetDT: Good point, removing the logline 👍

packages/ui-extensions-sdk/src/utils/security.ts
nmuldavin: Got it, thanks for the explanation!

PR #130, opened 2022-04-04
Run CI on `pull_request` events

joneshf-dd:
## Motivation

- We recently made some status checks required. This is what we wanted, but our current GitHub actions setup only works for `push` events. Since we have a GitHub Action that opens up PRs from outside the repo, we want the CI workflow to run on `pull_request` events. E.g. Look at https://github.com/DataDog/apps/pull/129. It isn't able to be merged because the CI check isn't run.
    Honestly, this was a mistake in our infrastructure that we never noticed up until now. So really, we're just fixing something that hasn't been working up until now.

## Changes

- We make the CI check run on both pushes and Pull Requests.

## Testing

- N/A

## Releases

Choose one:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: Continuous Integration
 
 on:
+    - pull_request
     - push
 
 jobs:


nmuldavin (approved):

joneshf-dd:
Thanks for the review!

PR #133, opened 2022-04-05
[UIAPPS-172] Implement Request Method to Fetch Custom Org Configs in DDClient

wfa207:
## Motivation

This PR implements a new `DDConfigClient` class that adds a new method to fetch custom app org configs.

## Changes

Surface `DDConfigClient` via `DDClient.config` to allow app developers to call `DDConfigClient.getConfig` method to fetch configuration values.

## Testing

This PR has automated tests.

## Releases

Choose one:

<details>
  <br />

  ```bash
  npm install @datadog/ui-extensions-sdk@0.31.2-canary.133.030bd14.0
  # or 
  yarn add @datadog/ui-extensions-sdk@0.31.2-canary.133.030bd14.0
  ```
</details>

### DIFF
--- a/.changeset/tender-buttons-refuse.md
+++ b/.changeset/tender-buttons-refuse.md
@@ -1,0 +1,5 @@
+---
+'@datadog/ui-extensions-sdk': minor
+---
+
+Implement Custom Config capabilities to `DDClient`
--- a/.changeset/tender-buttons-refuse.md
+++ b/.changeset/tender-buttons-refuse.md
@@ -1,5 +1,0 @@
----
-'@datadog/ui-extensions-sdk': minor
----
-
-Implement Custom Config capabilities to `DDClient`
--- a/packages/ui-extensions-sdk/src/config/config.ts
+++ b/packages/ui-extensions-sdk/src/config/config.ts
@@ -1,5 +1,10 @@
 import { RequestType } from '../constants';
-import { Config, ContextClient, LoggerClient, RequestClient } from '../types';
+import {
+    OrgConfig,
+    ContextClient,
+    LoggerClient,
+    RequestClient
+} from '../types';
 
 export class DDConfigClient {
     private readonly client: ContextClient & LoggerClient & RequestClient;
@@ -8,7 +13,7 @@ export class DDConfigClient {
         this.client = client;
     }
 
-    async getOrgConfig(): Promise<Config> {
+    async getOrgConfig(): Promise<OrgConfig> {
         return this.client.request(RequestType.GET_ORG_CONFIG);
     }
 }
--- a/packages/ui-extensions-sdk/src/config/config.ts
+++ b/packages/ui-extensions-sdk/src/config/config.ts
@@ -8,7 +8,7 @@ export class DDConfigClient {
         this.client = client;
     }
 
-    async getConfig(): Promise<Config> {
+    async getOrgConfig(): Promise<Config> {
         return this.client.request(RequestType.GET_CONFIG);
     }
 }
--- a/packages/ui-extensions-sdk/src/config/config.ts
+++ b/packages/ui-extensions-sdk/src/config/config.ts
@@ -9,6 +9,6 @@ export class DDConfigClient {
     }
 
     async getOrgConfig(): Promise<Config> {
-        return this.client.request(RequestType.GET_CONFIG);
+        return this.client.request(RequestType.GET_ORG_CONFIG);
     }
 }
--- a/packages/ui-extensions-sdk/src/config/config.ts
+++ b/packages/ui-extensions-sdk/src/config/config.ts
@@ -1,0 +1,14 @@
+import { RequestType } from '../constants';
+import { Config, ContextClient, LoggerClient, RequestClient } from '../types';
+
+export class DDConfigClient {
+    private readonly client: ContextClient & LoggerClient & RequestClient;
+
+    constructor(client: ContextClient & LoggerClient & RequestClient) {
+        this.client = client;
+    }
+
+    async getConfig(): Promise<Config> {
+        return this.client.request(RequestType.GET_CONFIG);
+    }
+}
--- a/packages/ui-extensions-sdk/src/types.ts
+++ b/packages/ui-extensions-sdk/src/types.ts
@@ -356,3 +356,7 @@ export interface LoadedResourceMetadata {
 export interface LoadedResourceMetaDataBatch {
     resources: LoadedResourceMetadata[];
 }
+
+export interface Config {
+    [key: string]: string | boolean;
+}
--- a/packages/ui-extensions-sdk/src/types.ts
+++ b/packages/ui-extensions-sdk/src/types.ts
@@ -357,6 +357,6 @@ export interface LoadedResourceMetaDataBatch {
     resources: LoadedResourceMetadata[];
 }
 
-export interface Config {
+export interface OrgConfig {
     [key: string]: string | boolean;
 }
--- a/packages/ui-extensions-sdk/src/utils/testUtils.ts
+++ b/packages/ui-extensions-sdk/src/utils/testUtils.ts
@@ -123,7 +123,9 @@ export class MockFramePostChildClient {
         );
     }
 
-    request(eventType: string, data: any): any {}
+    request(eventType: string, data: any): any {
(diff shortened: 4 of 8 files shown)


wfa207:
* master
  * **PR #133** <a href="https://app.graphite.dev/github/pr/DataDog/apps/133" target="_blank">Graphite</a>  (this PR)
    * **PR #135** <a href="https://app.graphite.dev/github/pr/DataDog/apps/135" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/apps/133?utm_source=stack-comment).

packages/ui-extensions-sdk/src/utils/testUtils.ts
```diff
@@ -123,7 +123,9 @@ export class MockFramePostChildClient {
         );
     }
 
-    request(eventType: string, data: any): any {}
+    request(eventType: string, data: any): any {
+        return this.mockRequest(eventType, data);
+    }
```
wfa207: Not sure if this was intentionally omitted, but by mocking the request automatically when `framepostClient.request` is called, we can test client methods that rely on the top-level `DDClient` classes `request` method.

packages/ui-extensions-sdk/src/config/config.ts
```diff
@@ -0,0 +1,14 @@
...
+export class DDConfigClient {
+    private readonly client: ContextClient & LoggerClient & RequestClient;
+
+    constructor(client: ContextClient & LoggerClient & RequestClient) {
+        this.client = client;
+    }
+
+    async getConfig(): Promise<Config> {
+        return this.client.request(RequestType.GET_CONFIG);
+    }
```
wfa207: I talked this over with @joneshf-dd — it feels best to implement this as a request since developers would likely want to request config values from `web-ui`. For full context, I briefly thought about packaging up config data in the `context`, but this would likely require better segmentation of the `context` object itself and dissociation between `framepost`'s `context` and the SDK Client's `context`.

packages/ui-extensions-sdk/src/utils/testUtils.ts
joneshf-dd: I think conceptually, it's not supposed to work, but it does because we're being lax with our testing.

If we think about the two sides of the system, we've got the Datadog platform and the SDK. The way things are supposed to work is that one side uses `request` and the other side uses `onRequest`. E.g. if the SDK uses `onRequest(RequestType.GET_WIDGET_CONTEXT_MENU_ITEMS, …)`, it's expected that the platform is going to use `request(RequestType.GET_WIDGET_CONTEXT_MENU_ITEMS, …)`; if the SDK uses `request(RequestType.SEND_NOTIFICATION, …)`, it's expected that the platform is going to use `onRequest(RequestType.SEND_NOTIFICATION, …)`.

What's not expected is that one side both sends a request and also responds to the same request. I'm not even sure it would actually do anything if you did that because the request is never routed to the sender always to the receiver.

Long story short, I think this is the reason the `request` method here had no implementation, because we haven't actually tested that when we end up calling `request` that it makes it across to the other side and returns something (because there is no other side in these tests).

I think what we do in other tests that make requests is setup the mock:
- https://github.com/DataDog/apps/blob/91301a66a276f39090db1f14c263e6a966e5edd9/packages/ui-extensions-sdk/src/notification/notification.test.ts#L23-L25
- https://github.com/DataDog/apps/blob/91301a66a276f39090db1f14c263e6a966e5edd9/packages/ui-extensions-sdk/src/api/api.test.ts#L13-L15
and validate that the right stuff is sent:
- https://github.com/DataDog/apps/blob/91301a66a276f39090db1f14c263e6a966e5edd9/packages/ui-extensions-sdk/src/notification/notification.test.ts#L32-L38
- https://github.com/DataDog/apps/blob/91301a66a276f39090db1f14c263e6a966e5edd9/packages/ui-extensions-sdk/src/api/api.test.ts#L26-L37

We don't seem to validate that the right stuff is returned.

.changeset/tender-buttons-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@datadog/ui-extensions-sdk': minor
```
joneshf-dd: I think the reason CI is failing is because we have to release all the the `@datadog/ui-extensions-*` packages in lock step so you can either [add an additional changeset](https://github.com/DataDog/apps/runs/5852676304?check_suite_focus=true#step:7:16), or add the react package to this changeset. Either way should get that step passing.

```suggestion
'@datadog/ui-extensions-react': minor
'@datadog/ui-extensions-sdk': minor
```

joneshf-dd (approved):
Looks good to me!

Tried to explain why CI is failing. Feel free to fix it however makes sense.

.changeset/tender-buttons-refuse.md
wfa207: Oh interesting; so, I guess, from what you're saying I just didn't pick all the right `ui-extensions-*` packages when I ran `yarn changeset add`?

packages/ui-extensions-sdk/src/utils/testUtils.ts
wfa207: Ahh I see. I think it makes more sense to validate what we're passing to the underlying methods rather than checking what's being returned. I'll update.

packages/ui-extensions-sdk/src/config/config.ts
nmuldavin: There's no doubt that the way you've done it is the simpler way, and I think it's right for now.

I see us almost certainly doing that work to dissociate framepost's `context` and the SDK's `context` anyways, at some point, in order to support a more rich data-model of information getting sent to the SDK. Hypothetically this could look something like:

```
{
  context: ...,
  orgConfig: ...,
  userConfig:..,
  state
}
```

but I don't know exactly what that will look like. We may even expose this data-structure directly from some hook like `client.getData()` or whatever. I talked with hardy about this and we agreed it's a good idea not to rush in to that before designing it carefully.

If I'm right that this is where we're heading, we'll have to go through a deprecation process of support for that request type down the line when we do ^ work, which will be very annoying, but at least we get to do that work after having defined the data-model more carefully.

nmuldavin: One thought though: 

What happens when if we end up doing arbitrary user-config later? Should we maybe name this `getOrgConfig()`? 

One general feedback about this feature cross-stack is that the naming feels inconsistent. We call it 'user settings' in the dev platform 'settings' in the installation modal, and 'config' here?

wfa207: I like the idea of renaming the method to `getOrgConfig`, and accordingly the request type, etc.

The naming is inconsistent across the stack, but I think each of the interfaces you mention carry slightly different context:
* In the Developer Platform, developers see "User Settings" since it's ultimately the user that will end up configuring these settings (I'm open to calling it "Org Settings", but I'm curious to hear Maureen's thoughts on it — if we decide Org Settings is better, then we can probably rename both the user-facing and developer-facing views the same)
* In my head, I view "configurations" as configured settings, whereas settings kind of just represents the universe of what can be toggled, so I'm okay with leaving it as "-config"

wfa207: Just recapping our offline conversation here — 

We're going with "Org Configuration" anytime developers are defining what's configurable and "Org Settings" anytime users are configuring individual settings.

packages/ui-extensions-sdk/src/types.ts
```diff
@@ -341,3 +341,7 @@ export interface DeprecatedEventUsage {
 }
 
 export type DeprecatedUsage = DeprecatedEventUsage; // we can union type here later
+
+export interface Config {
+    [key: string]: string | boolean;
+}
```
nmuldavin: Should we call this `OrgConfig` to be consistent + safe? It's likely user config would also be the same thing but we don't know for sure, we can always union-type later

wfa207: Ah oops; yeah — must've missed this one

wfa207: Alright, this should be updated

nmuldavin (approved):

PR #136, opened 2022-04-07
UIAPPS-166 Start widget context menu with 3rd-party data

joneshf-dd:
Motivation
----------

- We want to make a template for a widget context menu that interacts with 3rd-party data. This should continue to grow and change as features evolve.
- Jira Issue: https://datadoghq.atlassian.net/browse/UIAPPS-166

Changes
-------

- We create an example we can use as a template that solves a specific use case: an App where there's a widget context menu that interacts with 3rd-party data.
- We're following a lot of the setup from the custom widget with 3rd-party data from #125.

Testing
-------

Releases
--------

Choose one:


joneshf-dd:
* master
  * **PR #136** <a href="https://app.graphite.dev/github/pr/DataDog/apps/136" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/apps/136?utm_source=stack-comment).

examples/widget-context-menu-with-3rd-party-data/package.json
```diff
@@ -0,0 +1,48 @@
...
+    },
+    "scripts": {
+        "build": "react-scripts build",
+        "eject": "react-scripts eject",
+        "format:check": "prettier --check .",
+        "format:fix": "prettier --write .",
+        "lint:check": "eslint .",
+        "lint:fix": "yarn run lint:check --fix",
+        "lint": "yarn run lint:check && yarn run format:check",
+        "prepare": "exit 0",
```
wfa207, line 26: I thought this might get used by another `yarn` script, but I don't see this being used anywhere — what's this for?

examples/widget-context-menu-with-3rd-party-data/src/controller/widget-context-menu.ts
```diff
@@ -0,0 +1,130 @@
...
+                        await handleHostmapWidget(
+                            client,
+                            data.widgetInteraction.groupTags
+                        );
+                        break;
+
+                    /**
+                     * We're not handling clicks from any other types of widgets.
+                     * If we wanted to,
+                     * we would look at the other types available and process the data similarly.
```
wfa207, line 113: I think it's good that you're calling this out for the purposes of demonstrating how developers can extend functionality as needed.

👏👏👏

examples/widget-context-menu-with-3rd-party-data/src/controller/widget-context-menu.ts
```diff
@@ -0,0 +1,130 @@
...
+    );
+}
+
+/**
+ * We attempt to parse out the hostname from a `groupTag`.
+ * @param groupTag This should be in the form `key:value`.
+ * @returns The parsed hostname if it exists.
+ */
+function parseHostname(groupTag: string): string | undefined {
+    const [hostTag, hostName] = groupTag.split(':', 2);
```
wfa207, line 74: Are there cases where the `groupTag` can have more than one colon? If so, would we ever want the 3rd "tag" or some combination that **isn't** the first and second tag?

examples/widget-context-menu-with-3rd-party-data/src/host/host.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+}
+
+function isString(arg: unknown): arg is string {
+    return typeof arg === 'string';
+}
+
+/**
+ * We provide a helper for attempting to parse some arbitrary data into a {@link HostInformation} if possible.
+ * This is where we'd want to
```
wfa207: From the looks of this line and the broader file overall, I take it we're going to update this with more specifics at a later point?

wfa207 (approved):
This makes sense to me; I had a few questions, but I don't see why this can't be merged for the time being.

examples/widget-context-menu-with-3rd-party-data/package.json
joneshf-dd: **tl;dr;** This is so that the [top-level `"prepare"` script](https://github.com/DataDog/apps/blob/3e5e95858e6b3130a4ae6b1ba5389aed187121fd/package.json#L17) works. Because we say `yarn workspaces run prepare`, all of the workspaces need to have a `"prepare"` script, even if they do nothing except exit successfully.

---

The more in-depth answer is that there's a combination of things that cause this to happen:
- We're using yarn classic workspaces.
- We're using `yarn workspaces run`.
- We're developing in TypeScript.
- We're depending on other workspaces within this repo.

All of these things come together to mean that we need to figure out some way to build workspaces like `@datadog/ui-extensions-sdk` before attempting to do anything in a workspace like this example. The way we decided at the time was to lean on the `"prepare"` script since it runs before most other scripts in the lifecycle. Nothing requiring this to be the only way, and we have a Jira Issue to try and improve the situation: https://datadoghq.atlassian.net/browse/UIAPPS-135.

wfa207: Ah I see. Thanks for the background. I wasn't aware of the necessity of the `prepare` script.

examples/widget-context-menu-with-3rd-party-data/src/controller/widget-context-menu.ts
joneshf-dd: I honestly have no idea. I couldn't find a canonical explanation of what this syntax is. If you'd prefer, we can do something like this instead:

```TypeScript
function parseHostname(groupTag: string): string | undefined {
    if (!groupTag.startsWith('host:')) {
        return;
    }

    return groupTag.slice(5);
}
```

wfa207: I really don't have a preference; I was just curious. I think it's fine to leave it as-is.

examples/widget-context-menu-with-3rd-party-data/src/host/host.ts
joneshf-dd: This whole example is largely meant to be a starting point for App developers. The phrasing is meant to explain when to hook in, change something, expand the behavior, etc. We can for sure add to this later, but it's not really meant to be a fully completed thing.

If there's another way to explain things here, or a different set of things we should do, we can definitely do that.

joneshf-dd: Sorry, totally misunderstood. I dunno what this line was about, I'll remove it.

wfa207: All good!

joneshf-dd:
Thanks for the review!

PR #139, opened 2022-04-08
chore: remove VS Code Workspace Settings

ericrallen:
## Motivation

- Return Workspace-level Setting control to users

## Changes

- Remove extraneous `.vscode/settings.json` file and `.gitignore` it so developers have control over their own VS Code Workspace Settings

## Testing

- Open a TypeScript file in VS Code
- Make some changes to it
- Save and (if you have Format on Save enabled in your Settings) notice it formats properly based on our Prettier settings

## Releases

Choose one:


joneshf-dd (approved):
Thanks for doing this!

Issue #149, opened 2022-08-25
HandshakeTimeoutError: Handshake timed out


AbdallahHemdan:
I have followed the same steps mentioned [here in the documentation](https://docs.datadoghq.com/developers/datadog_apps/) for Datadog apps, But I am facing an issue in the `starter-kit` as following

- Run the `starter-kit`
- Click on `Click here to open your widget`
- When I try to click the `HERE` button I get the following error message

```
Uncaught (in promise) HandshakeTimeoutError: Handshake timed out
```
![image](https://user-images.githubusercontent.com/40190772/186642246-8e000846-f675-4532-b309-2f8f130da5ba.png)

- Node version: v14.17.0

AbdallahHemdan:
@rachelnicole It would be great if you could support here

rachelnicole:
@AbdallahHemdan can you provide the entire error with the stack trace / screenshot of the console. Also browser info

ericrallen:
@AbdallahHemdan Hey there!

You'll see this error message while viewing the widget locally because the Datadog Apps SDK cannot connect to the Datadog application.

These Datadog Apps are designed to run in an iframe inside of the Datadog UI ([described in more detail in this architecture document](https://github.com/DataDog/apps/blob/master/docs/en/programming-model.md)).

Did you see this error after following the instructions to create a new Developer Platform App and [adding your app to a Dashboard](https://docs.datadoghq.com/developers/datadog_apps/#add-your-app-to-a-dashboard)?

AbdallahHemdan:
@rachelnicole Thanks, I think @ericrallen commit to update the docs make sense now 🙏🏻

PR #150, opened 2022-08-30
docs: add note about handshaketimeout error

ericrallen:
## Motivation

This adds a note to the docs in reference to #149 to help prevent future confusion.

## Changes

Adds a note about the `HandshakeTimeoutError` in your Console when running the widget outside of a Datadog Dashboard.

Choose one:

### DIFF
--- a/docs/en/getting-started.md
+++ b/docs/en/getting-started.md
@@ -10,7 +10,7 @@ further_reading:
     text: "Programming Model"
 ---
 
-## Join the Beta! 
+## Join the Beta!
 Datadog Apps is currently in beta, but you can easily request access! [Use this form][5] to submit your request today. Once approved, you can start getting creative and develop your App for you, your organization, or for publishing to the entire Datadog community alongside our other great Datadog Apps!
 
 ## What is an App?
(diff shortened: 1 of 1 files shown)


docs/en/getting-started.md
```diff
@@ -47,6 +47,10 @@ Note that there are two pages:
 
 See the [Developer Platform Developer Guide][3] for details about this architecture.
 
+<div class="alert alert-info">
+You may notice an uncaught <strong>HandshakeTimeoutError</strong> in your JavaScript Console when you interact with the local widget in your browser directly. This is expected. As the <a href="https://github.com/DataDog/apps/blob/master/docs/en/programming-model.md">Developer Platform Developer Guide</a> explains that the Datadog Apps SDK connects to the Datadog User Interface via <a href="https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage">postMessage</a> and establishes a communication channel via the <a href="https://github.com/DataDog/apps/tree/master/packages/framepost"><code>@datadog/framepost</code> package</a> and the handshake attempt between the widget and the Datadog UI will timeout when there is no Datadog UI for the SDK has to communicate with.
```
wfa207: I don't think we need to call out that we're using `@datadog/framepost` for iFrame communication, since it's an implementation detail — what do you think?

wfa207 (approved):
This LGTM! I had a question about whether we care to be as specific about our implementation of channel communication, but it's not a blocker.

docs/en/getting-started.md
ericrallen: That's probably a good call.

I sort of just did a brain dump.

I'll take a pass at simplifying it.

PR #152, opened 2022-11-23
🐛 Build framepost for browser & node import

redking:
## Motivation

I'm currently migrating the Custom Widget component to the crosspage framework in [this PR](https://github.com/DataDog/web-ui/pull/72582).

As part of this work, custom widget was added to [the logic](https://github.com/DataDog/web-ui/pull/72582/files#diff-852608dd48bb80ddd6cb329ee4e86621df9910e37e8497091353f3c8108b319aR532) that calculates the widget title. To get a title for custom widgets, a new app manager is instantiated to retrieve the widget object. This code imports `framepost`. 

The same title generation code is also used in a set of [node unit tests](https://github.com/DataDog/web-ui/blob/preprod/services/mobile-widget-query/routes/get-computed-widgets/get-computed-widgets.unit.ts) related to the mobile platform (and, presumably, in the mobile code itself). Currently, these tests will fail at the point that `framepost` is imported with the error message 

```
ReferenceError: self is not defined

    > 1 | import { ClientDestroyedError } from '@datadog/framepost';
```

[You can see a failing build on the migration PR here](https://gitlab.ddbuild.io/DataDog/web-ui/-/jobs/196887726).

This happens because `framepost` is only built for the browser and not for consumption by node. 

## Changes

Updates the `globalObject` to be `this` rather than `self` (see [Webpack docs here](https://webpack.js.org/configuration/output/#outputglobalobject))

## Testing

I tested this locally, and the unit tests pass once this change is made.

The Webpack docs make it appear as though this has no negative side-effects for the browser release but, as I'm not totally familiar with how `framepost` is used, I'd appreciate some help testing.

## Releases

Choose one:

### DIFF
--- a/.changeset/witty-rats-sing.md
+++ b/.changeset/witty-rats-sing.md
@@ -1,0 +1,5 @@
+---
+'@datadog/framepost': minor
+---
+
+Build framepost for import by both the browser and node
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ coverage
 dist
 .DS_Store
 .vscode/settings.json
+.idea
--- a/packages/framepost/package.json
+++ b/packages/framepost/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@datadog/framepost",
-    "version": "0.3.1",
+    "version": "0.3.2",
     "description": "Secure parent-child iframe communication",
     "homepage": "https://github.com/DataDog/apps/tree/master/packages/framepost",
     "repository": {
--- a/packages/framepost/webpack.config.js
+++ b/packages/framepost/webpack.config.js
@@ -9,7 +9,8 @@ module.exports = {
         filename: 'framepost.min.js',
         path: path.resolve(__dirname, 'dist'),
         library: 'framepost',
-        libraryTarget: 'umd'
+        libraryTarget: 'umd',
+        globalObject: 'this'
     },
     module: {
         rules: [


enbashi (approved):

redking:
I'm closing this for the moment, as there's too much uncertainty on its potential impact. However, the PR can be reference for any future work to build these packages differently

nmuldavin:
I've been talking with @redking about this issue. I believe we should merge and cut a new release. More details here: https://github.com/DataDog/web-ui/pull/72582/files#r1040148873

The code causing the broken node test is the version of framepost that is imported directly by web-ui, so if we bump the version there it will fix the tests. Old versions of framepost may still run in the iframes, but we don't care about that. 

@joneshf-dd can you help Cormac do a release?

joneshf-dd:
Yeah, for sure. Full disclosure though, a lot of my knowledge here has gone away since we've not touched this code in a while.

It looks like the canary releases are bit-rotted at the moment. From what I remember [canary releases](https://github.com/DataDog/apps/blob/dfb8525d723358e1c2f78ba8af081fbcecdebacd/RELEASE.md#canary-releases) and [actual releases](https://github.com/DataDog/apps/blob/dfb8525d723358e1c2f78ba8af081fbcecdebacd/RELEASE.md#package-releases) follow different code paths. The regular releases might be fine, they might not. I honestly couldn't tell you without trying it out.

I'm assuming y'all don't want a canary release to test in `web-ui`. So the steps would be:
1. Merge this PR whenever you're ready.
2. See if `changeset` makes a PR for the normal release.
3. Merge the `changeset` PR to make the normal release.

If any of that doesn't work, we'll have to figure out why, or see if we can break glass and make a release without this process. Either way, merging this PR isn't going to make things worse, so start by merging this when you're ready.

redking:
OK, I'll merge this first thing tomorrow and we'll go from there

joneshf-dd:
Alright so whenever you're ready, merge https://github.com/DataDog/apps/pull/154 and with any luck, it'll cut a new release of `framepost` (and the other transitive dependencies).

joneshf-dd:
After a bit of figuring out stuff. We got it work. Looks like our `npm` credentials were invalidated. We got them refreshed, and [it works now](https://github.com/DataDog/apps/actions/runs/3647187331/jobs/6185566523#step:7:33) (at least the actual release worked).

Long story short, there's a [0.4.0 version of `framepost`](https://www.npmjs.com/package/@datadog/framepost/v/0.4.0) now that you can use. Lemme know if you need any more help.

redking:
@joneshf-dd Brilliant! Thanks for all your help!

